PR #33, opened 2015-01-23
Improvement on the integral contribution implementation

### LINKED ISSUE #32: Computation of i_term_
Before creating any pull request, I'd like to raise this for discussion.

Lines 282 and 315 in `pid.cpp`, explicitly:
```
i_term_ = i_term_ + gains.i_gain_ * dt.toSec() * p_error_;
```

Shouldn't this term be:
```
i_term_ = gains.i_gain_*( i_term_ +  dt.toSec() * p_error_ );
```

Note that, for instance, setting `gains.i_gain_` to zero online in the current implementation does not remove the integrated `i_term_` so far, and it won't until you reset the controller.

/cc @manuelbonilla

carlosjoserg:
The integral of the error is computed in a class member, and the integral contribution is computed as of it, instead of doing all together. Now, setting the integral gain to zero dynamically cancel the integral contribution.

Additional update:
Function `getCurrentPIDErrors(...)` now returns the integral of the position error, and not the contribution. Because of this, also the tests were modified. Tests `integrationWindupZeroGainTest` (actually, this one makes no sense now, since the error is directly given instead of obtaining by diving the term by the gain) and `integrationWindupTest` check the windup constraint on the computed command.

I don't know how this affects the dependant packages, but I believe the change is worthy.

Resolve #32.

### DIFF
--- a/include/control_toolbox/pid.h
+++ b/include/control_toolbox/pid.h
@@ -370,8 +370,8 @@ private:
 
   double p_error_last_; /**< _Save position state for derivative state calculation. */
   double p_error_; /**< Position error. */
-  double d_error_; /**< Derivative error. */
-  double i_term_;  /**< Integral term. */
+  double i_error_; /**< Integral of position error. */
+  double d_error_; /**< Derivative of position error. */
   double cmd_;     /**< Command to send. */
 
   // Dynamics reconfigure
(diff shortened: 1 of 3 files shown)


adolfo-rt:
LGTM.

On `getCurrentPIDErrors(...)` returning the integral error (correct since this patch) vs the integral term (incorrect, existing behavior), I don't know if the existing broken behavior is used by anyone.

I'm going to let this one soak another day to see if somebody wants to comment on this particular issue, otherwise I'll give it a go.

adolfo-rt:
Thanks for the two fixes!.

PR #37, opened 2015-05-07
Add optional state publishing to PID controller

paulbovbel:
Also a minor change to computeCommand calls, to chain the two implementations rather than replicate code.

### DIFF
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -124,6 +124,12 @@ bool Pid::init(const ros::NodeHandle &node, const bool quiet)
     gains.i_max_ = std::abs(gains.i_max_); // make sure the value is >= 0
   }
 
+  nh.param("publish_state", publish_state_, false);
+
+  if(publish_state_){
+    state_publisher_.init(nh, "state", 1);
+  }
+
   setGains(gains);
 
   reset();
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -305,7 +305,6 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
   if (dt == ros::Duration(0.0) || std::isnan(error) || std::isinf(error) || std::isnan(error_dot) || std::isinf(error_dot))
     return 0.0;
 
-
   // Calculate proportional contribution to command
   p_term = gains.p_gain_ * p_error_;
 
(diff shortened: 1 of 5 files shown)


adolfo-rt:
Per-changeset comments:

* 6dabd8f: In `control_msgs` there exists the [JointControllerState](https://github.com/ros-controls/control_msgs/blob/indigo-devel/control_msgs/msg/JointControllerState.msg), which is used in `ros_controllers` along with the `control_toolbox::Pid` class ([example](https://github.com/ros-controls/ros_controllers/blob/indigo-devel/effort_controllers/src/joint_position_controller.cpp#L216)). If possible, my suggestion would be to leverage this existing message type. Also, if you prefer to incorporate the PID state publishing inside the `control_toolbox::Pid` class, it would be desirable to also propagate this change to the few other controllers that use it.

* d68c920: Great to reduce code duplication!. I would suggest to make this a separate PR, as it's orthogonal to the above changeset.

* 31dd5e5: What's wrong with the newline at the end of the file?. This is a good UNIX practice.

paulbovbel:
6dabd8f: Thanks for pointing out that message, didn't spot it before. I'm a little skeptical of using it, since the PID library wouldn't always be used to control a joint, so the message name would be a little deceptive. It seems that in ROS the pattern is to create per-application message types for clarity. Additionally, from inside the controller I don't have access to the setpoint (only the error), so I can't fill that field.

paulbovbel:
Also, the JointStateController message doesn't include the three terms (p_term, i_term, d_term), which are the helpful bits for tuning a controller.

adolfo-rt:
I get that the `JointControllerState` name might be misleading, if you're using a PID to control a non-joint plant. I find it sensible to have a more generally named `PidState` message. There is in fact nothing joint-specific in the `JointControllerState` spec, and it kind of makes the assumptions that joint controllers are PID-based, which in general is not true.

I find the message type you propose less informative than `JointControllerState`. I get that the setpoint is not available from within the PID class, but if we're to propose a new message type, I'd really want it to convey as much useful information as possible.

Also, it would be best if new messages made it to `control_msgs`, and not to `control_toolbox`. The former has much lighter [dependencies](https://github.com/ros-controls/control_msgs/blob/indigo-devel/control_msgs/package.xml), and are all message related; while the latter brings more (and non-message related) [dependencies](https://github.com/ros-controls/control_toolbox/blob/indigo-devel/package.xml).

paulbovbel:
Sounds good on all points. Any proposals for what additional information you'd like to see in the PidState message?

paulbovbel:
Moved message to control_msgs, depends on https://github.com/ros-controls/control_msgs/pull/14

paulbovbel:
Depends on #44 or #42

paulbovbel:
Updated with https://github.com/ros-controls/control_msgs/pull/14, other dependencies met, ping for review.

adolfo-rt:
Please rebase now that #46 has been merged, and we preserved the implementation in `computeCommand`.

Also, please make two commits: One with the style change, and another one with the enhancement.

paulbovbel:
Ping for review. No style change here.

src/pid.cpp
```diff
@@ -318,6 +323,25 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
   // Compute the command
   cmd_ = p_term + i_term + d_term;
 
+  // Publish controller state if configured
+  if (publish_state_ && state_publisher_.trylock())
+  {
+    state_publisher_.msg_.header.stamp = ros::Time::now();
```
adolfo-rt: Hmmm, I realize just now that under certain RTOSes, like Xenomai 2.6.X, `ros::Time::now` is a no-no, as there's a Linux system call involved.

Since state publishing is optional and comes disabled by default, I think your proposal is an acceptable compromise.

Could you please document the `publish_state` parameter [here](https://github.com/ros-controls/control_toolbox/blob/indigo-devel/include/control_toolbox/pid.h#L81), and mention that if enabled, it may break real-time guarantees on certain RTOS?. I think that with this, the PR can be merged.

paulbovbel:
@adolfo-rt, added the documentation, thanks

bmagyar:
Could you update this please?

paulbovbel:
Rebased on kinetic-devel

bmagyar:
Unfortunately you have to open a new PR pointing to kinetic-devel, github is rigid on the "from-to" relations of PRs, cannot  edit this one...

PR #39, opened 2015-05-08
Update computeCommand call to match documented/expected behaviour

paulbovbel:
computeCommand is documented to use 'error = target - state' rather than the old 'pError = pState-pTarget' of updatePid. One of the computeCommand implementations behaves correctly (https://github.com/ros-controls/control_toolbox/blob/indigo-devel/src/pid.cpp#L372), but this one does not (https://github.com/ros-controls/control_toolbox/blob/indigo-devel/src/pid.cpp#L372).

Next pull will be to chain the calls and prevent this kind of divergence.

### DIFF
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -369,7 +369,7 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
   d_term = gains.d_gain_ * d_error_;
 
   // Compute the command
-  cmd_ = - p_term - i_term - d_term;
+  cmd_ = p_term + i_term + d_term;
 
   return cmd_;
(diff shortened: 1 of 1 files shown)


adolfo-rt:
Good stuff @paulbovbel. Could you please add a few test cases to the unit test suite, so the three-parameter version of `computeCommand` code is exercised?. The two-parameter version of the method is already tested (and works). The old saying speaks the truth:

'if it's untested, it's broken'.

adolfo-rt:
Merging because this is affecting people. Please consider extending the test suite so we have a regression test.

@carlosjoserg

Issue #41, opened 2015-05-22
Increase coverage of PID class test suite


adolfo-rt:
...to prevent situations such as those reported in #40.

@paulbovbel, @carlosjoserg

christophfroehlich:
PID has now almost 95% test coverage
https://app.codecov.io/gh/ros-controls/control_toolbox/blob/ros2-master/src%2Fpid.cpp

Issue #40, opened 2015-05-22
DANGEROUS: Wrong behavior at startup after 1.13.0 -> 1.13.1 upgrade


marcoesposito1988:
After upgrading today I started my robot (KUKA LWR4+ with effort joint_trajectory_controller and MoveIt) and it immediately started going at full speed without any reason. I could observe this behavior consistently until I downgraded control_toolbox to 1.13.0 (it tried to reach the same position again and again at full speed, without any care for the joint limits).

This is very dangerous (I had my hand on the red button by pure chance, otherwise I would have damaged my robot very heavily).

Am I the only one with this problem? 

@carlosjoserg, I see there is a commit of yours there. Didn't you have any problems with this? Strange, since we have the same setup.

adolfo-rt:
Thanks for reporting. I believe to have found the culprit. A sign was [mistakenly flipped](https://github.com/ros-controls/control_toolbox/commit/bc45970b685c3cfdb33672897a3eec090b28ee71?diff=unified#commitcomment-11329517) in bc45970b685c3cfdb33672897a3eec090b28ee71. I believe you must be using this method.

A fix was proposed in #39. I'll roll out 1.13.2 right away, and a communication to the relevant mailing lists.

adolfo-rt:
1.13.2 released into [indigo](https://github.com/ros/rosdistro/pull/8520) and [jade](https://github.com/ros/rosdistro/pull/8521).

marcoesposito1988:
Bingo! It works now. Thanks a lot for the quick response!

adolfo-rt:
Thanks for confirming the fix.

carlosjoserg:
> Didn't you have any problems with this? Strange, since we have the same setup.
No, I was still using 1.13.0 from synaptics.

Certainly quick, I couldn't even blink!

rctoris:
@tfoote Any chance we can get a new sync given this critical bug now in the debians?

tfoote:
Yeah, once things have finished building. The farm has a queue of ~ 3700 packages to build at the moment.

PR #42, opened 2015-05-25
Response to #40

carlosjoserg:
Could be that #40 happened mainly due to 2 things (well, 3, if you count my awkwardness):

1. Lack of tests on command computation, as already requested by @adolfo-rt in #41
2. The `computeCommand` method has two signatures with almost a replica of code within -> Same code in two places requires update in two places -> The chance of making a mistake when updating increases.

In this initial PR both are addressed, there are 2 separated commits to easy the review.

First, in 9dc84fd548c0bd7df9d428cf60594fde24ea2b6d, the coverage of the PID class test suite is increased with the following modification of current tests:

* `zeroITermBadIBoundsTest`: I think that entering `i_max < i_min` should be prevented, or at least warned. A suggested change to avoid bad i-bounds is there commented because it would require a change in the `setGains` method return type to `bool` instead of `void` and the current implementation wouldn't pass the test. Another posibility would be to leave that return type and just throw a warning if `i_max < i_min` is passed.

* `integrationWindupTest`: Now upper and lower bounds are tested.

* integrationWindupZeroGainTest: Untouched. However, it looks to me it is useless with the current integral contribution implementation.

* `gainSettingCopyPIDTest`: Makeup in names for readability + Added test on `reset()` method.

And the following additions, a new set of tests under the name `CommandTest`, hopefully well described with comments:

* `proportionalOnlyTest`, `integralOnlyTest`, `derivativeOnlyTest`: Check that the command computed using each contribution independently is correct. Note that, they depend on the integral and differentiation schemes, and if they change, the tests might not be valid anymore.

* `completePIDTest`: Few checks on the sum up of contributions. Same here, the test depends on the integral and differentiation schemes.

Second, in 7fe5e126b05c4c915ecbb4c143ffef6b8c0ff5c1, only one method does the PID command computation and writes to class members (except for `p_error_last_`), which is the one with the signature that allows users to pass a pre-computed error derivative. The other one calls this method passing the error derivative computed with an internal differentiation scheme. I couldn't find whether this is ROS-style compliant or not, and not sure if it affects critically the performance, but surely it will reduce the chance of making a mistake if the implementation is to be updated again.

PS: I didn't do anything with the DEPRECATED command computation methods.

### DIFF
--- a/test/pid_tests.cpp
+++ b/test/pid_tests.cpp
@@ -9,17 +9,16 @@
 
 using namespace control_toolbox;
 
-TEST(ParameterTest, zeroITermBadIBoundsTest)
+TEST(ParameterTest, ITermBadIBoundsTest)
 {
-  RecordProperty("description","This test checks robustness against divide-by-zero errors when given integral term bounds which do not include 0.0.");
+  RecordProperty("description","This test checks that the integral contribution is robust to bad i_bounds specification (i.e. i_min > i_max).");
 
-  Pid pid(1.0, 0.0, 0.0, -1.0, 0.0);
-
+  // Check that the output is not a non-sense if i-bounds are bad, i.e. i_min > i_max
+  Pid pid(1.0, 1.0, 1.0, -1.0, 1.0);
   double cmd = 0.0;
   double pe,ie,de;
 
   cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
-
   pid.getCurrentPIDErrors(&pe,&ie,&de);
   EXPECT_FALSE(boost::math::isinf(ie));
   EXPECT_FALSE(boost::math::isnan(cmd));
@@ -34,16 +33,18 @@ TEST(ParameterTest, integrationWindupTest)
 {
   RecordProperty("description","This test succeeds if the integral contribution is prevented from winding up when the integral gain is non-zero.");
 
-  Pid pid(0.0, 2.0, 0.0, 1.0, -1.0);
+  Pid pid(0.0, 1.0, 0.0, 1.0, -1.0);
 
   double cmd = 0.0;
   double pe,ie,de;
 
-  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
+  // Test lower limit
+  cmd = pid.computeCommand(-10.03, ros::Duration(1.0));
   EXPECT_EQ(-1.0, cmd);
 
-  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
-  EXPECT_EQ(-1.0, cmd);
+  // Test upper limit
+  cmd = pid.computeCommand(30.0, ros::Duration(1.0));
+  EXPECT_EQ(1.0, cmd);
 }
 
 TEST(ParameterTest, integrationWindupZeroGainTest)
@@ -133,11 +134,11 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
   EXPECT_EQ(i_min, i_min_return);
 
   // Test that errors are zero
-  double pe, ie, de;
-  pid2.getCurrentPIDErrors(&pe, &ie, &de);
-  EXPECT_EQ(0.0, pe);
-  EXPECT_EQ(0.0, ie);
-  EXPECT_EQ(0.0, de);
+  double pe2, ie2, de2;
+  pid2.getCurrentPIDErrors(&pe2, &ie2, &de2);
+  EXPECT_EQ(0.0, pe2);
+  EXPECT_EQ(0.0, ie2);
+  EXPECT_EQ(0.0, de2);
     
   // Test assignment constructor -------------------------------------------------
   Pid pid3;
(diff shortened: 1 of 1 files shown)


carlosjoserg:
To make issue 2 in this PR clearer, check that, in #38, only one of the `computeCommand` methods is updated. So, if someone uses the other signature, the added feature wouldn't be available.

paulbovbel:
+1, this supercedes #44, which can be closed if this is merged

test/pid_tests.cpp
```diff
@@ -11,15 +11,22 @@ using namespace control_toolbox;
 
 TEST(ParameterTest, zeroITermBadIBoundsTest)
 {
-  RecordProperty("description","This test checks robustness against divide-by-zero errors when given integral term bounds which do not include 0.0.");
+  RecordProperty("description","This test checks that the limits for the integral contribution are entered correctly.");
 
-  Pid pid(1.0, 0.0, 0.0, -1.0, 0.0);
+  //// Do not allow this construction
```
adolfo-rt: Please either remove dead code, or bring it to a functional state.

adolfo-rt:
Could you please rebase against `indigo-devel`. I think it should be straightforward (or fast-forward, to be more precise ;-).

Many thanks for the extra tests.

carlosjoserg:
Sure. 

I'll also delete 7fe5e126b05c4c915ecbb4c143ffef6b8c0ff5c1 since #44 was already merged, it was exactly the same.

carlosjoserg:
[update]
- Rebased to `indigo-devel` (all tests passed)
- Minor, changed the name of `zeroITermBadIBoundsTest` to `ITermBadIBoundsTest`
- Deleted 7fe5e126b05c4c915ecbb4c143ffef6b8c0ff5c1 from PR

bmagyar:
I'm about to press some green buttons here. It seems that this PR got a bit forgotten, please raise a thumb if you also think it should be merged.

test/pid_tests.cpp
```diff
@@ -152,11 +153,143 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
+  double cmd = 0.0;
+
+  // If initial error = 0, i-gain = 1, dt = 1
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect command = error
+  EXPECT_EQ(-0.5, cmd);
+
+  // If call again with same arguments
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect the integration do it part doubling the command
```
efernandez: `Then expect the integral part doubles the command`

test/pid_tests.cpp
```diff
@@ -152,11 +153,143 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
+
+  // All contributions are tested, here few tests check that they sum up correctly
+  // If initial error = 0, all gains = 1, dt = 1
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect command = 3x error
+  EXPECT_EQ(-1.5, cmd);
+
+  // If call again with same arguments, no error change, but integration do its part
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect command = 3x error againa
```
efernandez: `again` (not `againa`)

test/pid_tests.cpp
```diff
@@ -152,11 +153,143 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect the integration do it part doubling the command
+  EXPECT_EQ(-1.0, cmd);
+
+  // Call again with no error
+  cmd = pid.computeCommand(0.0, ros::Duration(1.0));
+  // Expect the integral part to keep the previous command because ensure error = 0
+  EXPECT_EQ(-1.0, cmd);
+
+  // Double check that the integral contribution keep the previous command
```
efernandez: `keeps`

test/pid_tests.cpp
```diff
@@ -152,11 +153,143 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
+  EXPECT_EQ(-0.5, cmd);
+
+  // If call again with same arguments
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect the integration do it part doubling the command
+  EXPECT_EQ(-1.0, cmd);
+
+  // Call again with no error
+  cmd = pid.computeCommand(0.0, ros::Duration(1.0));
+  // Expect the integral part to keep the previous command because ensure error = 0
```
efernandez: `it ensures`

test/pid_tests.cpp
efernandez: or
`Then expect the integration do its part doubling the command`

test/pid_tests.cpp
```diff
@@ -152,11 +153,143 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
+  EXPECT_EQ(-0.5, cmd);
+
+  // If call again
+  cmd = pid.computeCommand(-0.5, ros::Duration(1.0));
+  // Then expect the same as before
+  EXPECT_EQ(-0.5, cmd);
+
+  // If call again doubling the error
+  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
+  // Then expect the command doubl-ed
```
efernandez: `doubled`

efernandez:
LGTM, but I've added a few comments regarding typos on the comments. @bmagyar Maybe you can fix them before merging.

All the tests look good. Thanks for the contribution.

bmagyar:
Well I don't really have access to Carlos' branch, but I can fix it after I merged it! :)

Oh btw, merging...

bmagyar:
Oh shoot, should have rebased this one...

carlosjoserg:
Hey, sorry I'm late for the party... thanks for the review and the merge!

bmagyar:
No problem, thanks for your contribution!

PR #43, opened 2015-05-31
Integral windup

mehdierc:
i think we should limit i_error_ for preventing i_term to go out of bounds, if we just limit i_term, i_error can increase over time and cause a lag to PID respond. :)


paulbovbel:
This is addressed in #38, in addition to adding a param to toggle the antiwindup behaviour.

bmagyar:
Thanks for the contribution, we have addressed this in the ROS Kinetic version!
I'm closing this PR now, feel free to open a new issue or PR if you have more to add to Kinetic.

PR #45, opened 2015-07-01
Added service for reading PIDs as well pid initialisation with weight…

cehberlin:
…s of first pid controller if several controllers are added to the pid gain setter in order to mimic the idea of setting simlar weights to all controllers like done in setGains

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog for package control_toolbox
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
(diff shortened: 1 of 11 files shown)


adolfo-rt:
Instead of a ROS service, would a topic broadcasting PID state work for you as well?. Take a look at https://github.com/ros-controls/control_msgs/pull/14. The message is quite complete, and exposes PID gains, among other things.

cehberlin:
I my case as well as in general I think it is much more useful to get this information on request, since it is not data that is frequently changed. Moreover, I would like to have it directly instead of waiting for a topic message.

christophfroehlich:
Not relevant for ROS 2 any more

PR #46, opened 2015-08-11
Remove code repetition

paulbovbel:
Remove code repetition by collapsing deprecated computeCommand into negative updatePid function call

### DIFF
--- a/include/control_toolbox/pid.h
+++ b/include/control_toolbox/pid.h
@@ -117,10 +117,10 @@ public:
   /*!
    * \brief Store gains in a struct to allow easier realtime buffer usage
    */
-  struct Gains 
+  struct Gains
   {
     // Optional constructor for passing in values
-    Gains(double p, double i, double d, double i_max, double i_min) 
+    Gains(double p, double i, double d, double i_max, double i_min)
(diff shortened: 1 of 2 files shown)


paulbovbel:
Trailing whitespace got trimmed, see https://github.com/ros-controls/control_toolbox/pull/46/files?w=1

adolfo-rt:
Although practically the effect would be the same right now, I would strongly recommend providing the patch the other way around: keep the details in `computeCommand`, and implement `updatePid` in terms of `-computeCommand`. The reason being that [`updatePid`](https://github.com/ros-controls/control_toolbox/blob/indigo-devel/include/control_toolbox/pid.h#L305-L321) is deprecated, and might go away some day. It'll be much easier to remove it if it does not contains the actual implementation.

Also, I'm fine with the whitespace fix, but prefer it in a separate commit. I would suggest to first apply that change on the vanilla file, make a style commit, and then proceed to apply the actual code changes.

paulbovbel:
Ah, mixed up which one is deprecated, thanks

adolfo-rt:
Thanks @paulbovbel. Ready to merge!

PR #47, opened 2015-08-13
Address -Wunused-parameter warnings

adolfo-rt:


adolfo-rt:
No news, good news. Merging.

PR #48, opened 2015-10-09
Added feed-forward tool.

praij90:
Added a universal feed forward tool.

Accepts viscous friction and Columb friction (can be different for positive and negative side).


praij90:
I have made other additions as well so I will close this pull request. I will make a new pull(/merge) request which has the additional features as well.

-Paul

PR #49, opened 2015-10-09
Additional features for the control_toolbox

praij90:
Added a generic feed forward tool and a block wave signal source tool.
- Feed forward: This implements a simple viscous and Columb friction model to compute the command effort. This can be used in addition to the PID tool in several controllers.
- Block wave: This tool generates a block wave signal that can be used for the feed forward identification procedure. Moreover it can be used to test a system by essentially being equivalent to the application of successive steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ include_directories(
   )
 
 add_library(${PROJECT_NAME}
+  src/ff.cpp
   src/pid.cpp
   src/pid_gains_setter.cpp
   src/sine_sweep.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ include_directories(
   )
 
 add_library(${PROJECT_NAME}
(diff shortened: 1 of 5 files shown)


praij90:
Hello ROSers,

I have made two additions to the toolbox. I have used them in my project and they seem very useful. If you agree I would be honoured if you were to include them in your repository.

Happy coding,
 -Paul

adolfo-rt:
Hello, @praij90. We're currently a bit slow processing pending PRs, so I appreciate your patience. I'd like to point out a few things to take into account if you want to make your work available to others:

* * Use descriptive names. `class Ff` doesn't say anything about what it does. Something along the lines of  `FrictionCompensator`, or something even more descriptive would be much better.

* Document the API. What do classes/methods do, pre/post conditions and invariants (if any). Currently, if you don't read the code, you don't know what things are and how to use them.

* Adhere to the code style of the project. Naming conventions, indentation, etc.

PR #50, opened 2016-01-22
changed the range of dynamic reconfigure to allow negative ones

VahidAminZ:
dynamic reconfigure only allows setting positive values for PID gains although negative values can be set using setGain.

### DIFF
--- a/cfg/Parameters.cfg
+++ b/cfg/Parameters.cfg
@@ -40,9 +40,9 @@ PACKAGE='control_toolbox'
 
 def generate(gen):
 	#        Name            Type    Level  Description      Default  Min  Max
-	gen.add( "p" ,      double_t, 1,"Proportional gain.", 10.0 , 0 , 100000)
-	gen.add( "i" ,      double_t, 1,"Integral gain.",     0.1 , 0 , 1000)
-	gen.add( "d" ,      double_t, 1,"Derivative gain.",   1.0 , 0 , 1000)
+	gen.add( "p" ,      double_t, 1,"Proportional gain.", 10.0 , -100000 , 100000)
+	gen.add( "i" ,      double_t, 1,"Integral gain.",     0.1 , -1000 , 1000)
+	gen.add( "d" ,      double_t, 1,"Derivative gain.",   1.0 , -1000 , 1000)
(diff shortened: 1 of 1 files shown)


bmagyar:
:+1:

cfg/Parameters.cfg
```diff
@@ -40,9 +40,9 @@ PACKAGE='control_toolbox'
 
 def generate(gen):
 	#        Name            Type    Level  Description      Default  Min  Max
-	gen.add( "p" ,      double_t, 1,"Proportional gain.", 10.0 , 0 , 100000)
-	gen.add( "i" ,      double_t, 1,"Integral gain.",     0.1 , 0 , 1000)
-	gen.add( "d" ,      double_t, 1,"Derivative gain.",   1.0 , 0 , 1000)
+	gen.add( "p" ,      double_t, 1,"Proportional gain.", 10.0 , -100000 , 100000)
```
efernandez: Just wondering if it makes sense to have a negative proportional gain wrt to the error.
I'd say the same for the integral gain.
AFAIK, only the derivative gain can be negative in a stable PID controller.

Maybe you're computing your error as actual - desired (setpoint) in your controller?!

If there's a case where negative values for the proportional and integral gains makes sense, could you provide it?
For the derivative the condition for a stable PID controller could still allow for negative values though.

http://www.mathworks.com/help/slcontrol/ug/troubleshooting-automatic-pid-tuning.html

VahidAminZ: hey @efernandez. I can think of the following system: G(s) = -1/s-3 which is unstable.
Adding a P controller results in transfer function of -p/(s-3-p) and this becomes stable for p < -3. Also if you want to add an integral part it should be negative. Generally p and I should have same sign.

bmagyar: I have to admit I also did a bit of homework on this before putting the +1 :)

Allow me to rephrase the situation bluntly. The PID class lets you set negative gains for all, so it is best to keep this consistent with the behaviour of the `dynamic_reconfigure` window. This PR itself should be that simple.

It does however remind us that there is no test on negative gains in the class, and question is if it makes sense to implement some for that later. (Not the scope of this PR)

efernandez: Fair enough :+1:

bmagyar:
Merging...

PR #56, opened 2016-05-19
Fix print statement in Parameters.cfg to be compatible with Python3

igor-nap:


bmagyar:
Hi Igor,
Thanks for the contribution. I gave it a quick try, but I'm not entirely satisfied with the results.

Running the print statement on Python2.7 gave me the following:
```
('ERROR', 'hello_from_the_Dark_side', 'Parameters.cfg failed using parameter_generator_catkin, using rosbuild instead')
```
while Python3 prints
```
ERROR hello_from_the_dark_side Parameters.cfg failed using parameter_generator_catkin, using rosbuild instead
```

According to the Kinetic (REP)[http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021] both Python2 and Python3 are supported.


BUT I'm actually not sure we need that part in the `Package.cfg` at all, seems like some leftover from rosbuild support which we're not doing anymore. Could you try to remove that error handling and convert the file to simply follow what's on the (tutorial)[http://wiki.ros.org/dynamic_reconfigure/Tutorials/HowToWriteYourFirstCfgFile] please?

igor-nap:
Thanks for the suggestion, Bence! I've made a new [PR](github.com/ros-controls/control_toolbox/pull/57) with the alternative solution. Let me know what you think.

PR #65, opened 2017-11-08
Add control_msgs to CATKIN_DEPENDS.

mikepurvis:
Needed because of the usage here:

https://github.com/ros-controls/control_toolbox/blob/ba1100638a48a461e4e642150f4ae666303e3ecc/include/control_toolbox/pid.h#L40


bmagyar:
The travis build on this one is broken, we'll see CI results once merged.

PR #63, opened 2017-03-15
Updated travis configuration using industrial_ci

progtologist:
Please update the recipients accordingly.
Also refers to issue #61

PR #66, opened 2017-11-08
switched to industrial_ci

### LINKED ISSUE #61: Travis build job needs update
Currently all the branches in the control_toolbox still use the old configuration (without docker) that means that all building is done using hydro (even for the kinetic branch).

Although it works at the moment, we really should update it as various issues might arise and go unnoticed (all builds are tested using ubuntu 12.04 which is quite old).

I can make a PR with a fix if the maintainers agree as well.

ipa-mdl:
supersedes #63 
closes #61 

same as https://github.com/ros-controls/realtime_tools/pull/20

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,48 +1,28 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
-# Author: Dave Coleman
-language:
-  - cpp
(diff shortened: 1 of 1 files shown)


bmagyar:
Thank you!

PR #60, opened 2017-03-13
Added LeadLag controller

progtologist:
This PR should be able to be integrated in any version(branch), I am currently using indigo so I have only tested it against that.

PR #68, opened 2017-11-30
Added LeadLag controller

progtologist:
Continues #60


src/pid.cpp
```diff
@@ -400,9 +400,6 @@ double Pid::getCurrentCmd()
 
 void Pid::getCurrentPIDErrors(double *pe, double *ie, double *de)
 {
-  // Get the gain parameters from the realtime buffer
-  Gains gains = *gains_buffer_.readFromRT();
```
bmagyar: I remember this frmo the previous PR. Reading the code suggests that really nothing happens here, right?

Could you please put this change into a different commit? It doesn't belong with the LeadLag stuff. Could still pass within the same PR as a minor fix.

src/lead_lag.cpp
```diff
@@ -0,0 +1,245 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/*
+  Author: Aris Synodinos
+  Desc: Implements a standard lead-lag controller
+*/
+
+#include <control_toolbox/lead_lag.h>
+#include <tinyxml.h>
```
bmagyar: Do we really need tinyxml here? Could you check if the PID class also needs it? I had a look today and it didn't seem so...

bmagyar:
Don't worry about the test that's failing now, it's due to our source-based test that lacks dependencies on Lunar.

src/lead_lag.cpp
progtologist: Yes, I don't really know if anyone uses it, but it is part of the public API for the PID class as well here
``` cpp
bool initXml(TiXmlElement *config);
```

src/pid.cpp
progtologist: Fixed ;)

christophfroehlich:
Needs to be ported to ROS 2

PR #71, opened 2019-01-14
fixed an antiwindup bug

onionsflying:
If gains.i_gain_ < 0 and std::abs(gains.i_min_) != std::abs(gains.i_max_), there would be a bug.  
For example: i_error_ = 0.1, gains.i_gain_ = -2.5, gains.i_min_ = -0.2, gains.i_max_ = 0.5.  
Then the calculated i_term would be -0.25, which is out of the range [-0.2, 0.5]

### DIFF
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -42,6 +42,7 @@
 #include <tinyxml.h>
 
 #include <boost/algorithm/clamp.hpp>
+#include <boost/algorithm/minmax.hpp>
 
 namespace control_toolbox {
 
@@ -335,12 +336,11 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
   // Calculate the integral of the position error
   i_error_ += dt.toSec() * p_error_;
 
-  if(gains.antiwindup_)
+  if(gains.antiwindup_ && gains.i_gain_!=0)
   {
     // Prevent i_error_ from climbing higher than permitted by i_max_/i_min_
-    i_error_ = boost::algorithm::clamp(i_error_,
-                                       gains.i_min_ / std::abs(gains.i_gain_),
-                                       gains.i_max_ / std::abs(gains.i_gain_));
+    boost::tuple<double, double> bounds = boost::minmax<double>(gains.i_min_ / gains.i_gain_, gains.i_max_ / gains.i_gain_);
+    i_error_ = boost::algorithm::clamp(i_error_, bounds.get<0>(), bounds.get<1>());
   }
 
   // Calculate integral contribution to command
--- a/test/pid_tests.cpp
+++ b/test/pid_tests.cpp
@@ -100,25 +100,25 @@ TEST(ParameterTest, negativeIntegrationAntiwindupTest)
 {
   RecordProperty("description","This test succeeds if the integral error is prevented from winding up when i_gain < 0");
 
-  double i_gain = -2.0;
-  double i_min = -1.0;
-  double i_max = 1.0;
+  double i_gain = -2.5;
+  double i_min = -0.2;
+  double i_max = 0.5;
   Pid pid(0.0, i_gain, 0.0, i_max, i_min, true);
 
   double cmd = 0.0;
   double pe,ie,de;
 
-  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
-  EXPECT_EQ(1.0, cmd);
+  cmd = pid.computeCommand(0.1, ros::Duration(1.0));
+  EXPECT_EQ(-0.2, cmd);
 
-  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
-  EXPECT_EQ(1.0, cmd);
+  cmd = pid.computeCommand(0.1, ros::Duration(1.0));
+  EXPECT_EQ(-0.2, cmd);
 
-  cmd = pid.computeCommand(0.5, ros::Duration(1.0));
-  EXPECT_EQ(0.0, cmd);
+  cmd = pid.computeCommand(-0.05, ros::Duration(1.0));
+  EXPECT_EQ(-0.075, cmd);
 
-  cmd = pid.computeCommand(-1.0, ros::Duration(1.0));
-  EXPECT_EQ(1.0, cmd);
+  cmd = pid.computeCommand(0.1, ros::Duration(1.0));
+  EXPECT_EQ(-0.2, cmd);
 }
 
 TEST(ParameterTest, gainSettingCopyPIDTest)


bmagyar:
@ipa-mdl would you mind taking a look at this?

src/pid.cpp
```diff
@@ -335,21 +335,16 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
...
-  if(!gains.antiwindup_)
+  // Limit i_term so that the limit is meaningful in the output
+  i_term = boost::algorithm::clamp(i_term, gains.i_min_, gains.i_max_);
+
+  if(gains.antiwindup_ && gains.i_gain_!=0)
   {
-    // Limit i_term so that the limit is meaningful in the output
-    i_term = boost::algorithm::clamp(i_term, gains.i_min_, gains.i_max_);
+    // Prevent i_error_ from climbing higher than permitted by i_max_/i_min_
+    i_error_ = i_term / gains.i_gain_;
```
ipa-mdl: this does change the value every time..

I'd like to use the fixed clamping instead (not yet tested):
```c++
std::pair<double> bounds = std::minmax(gains.i_min_ / gains.i_gain_, gains.i_max_);
i_error_ = boost::algorithm::clamp(i_error_, bounds.first, bound.second);	
```
And please place it at the old location, this way there is no disparity

onionsflying: That seems great, but `std::minmax` may be a C++11 function. I can't use it unless I change CMakeLists.txt. So I used `std::min` and `std::max`.

onionsflying:
I created a new branch [kinetic-patch](https://github.com/onionsflying/control_toolbox/tree/kinetic-patch). We could use this new branch if you think that the first commit should be adding a test.

src/pid.cpp
```diff
@@ -335,12 +336,13 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
...
   i_error_ += dt.toSec() * p_error_;
 
-  if(gains.antiwindup_)
+  if(gains.antiwindup_ && gains.i_gain_!=0)
   {
     // Prevent i_error_ from climbing higher than permitted by i_max_/i_min_
-    i_error_ = boost::algorithm::clamp(i_error_,
-                                       gains.i_min_ / std::abs(gains.i_gain_),
-                                       gains.i_max_ / std::abs(gains.i_gain_));
+    double i_error_min = std::min<double>(gains.i_min_ / gains.i_gain_, gains.i_max_ / gains.i_gain_);
```
ipa-mdl: Since we are using  `boost::algorithm`  for the clamping, we could use `<boost/algorithm/minmax.hpp>` as well.
I'd like to avoid this duplication..

test/pid_tests.cpp
```diff
@@ -121,6 +121,31 @@ TEST(ParameterTest, negativeIntegrationAntiwindupTest)
   EXPECT_EQ(1.0, cmd);
 }
 
+TEST(ParameterTest, negativeAsymmetricIntegrationAntiwindupTest)
```
ipa-mdl: IMHO we don't need another test. Just update the values in negativeIntegrationAntiwindupTest :)

onionsflying:
Hi @ipa-mdl , would you mind taking a look at these two commits?

ipa-mdl:
code looks good, but I pushed the commits individually to see if the test fails without the fix

ipa-mdl:
For some reason the test did pass please have a look at it again.

ipa-mdl:
I guess you have to build up the error first.

onionsflying:
I've tested these two commits on my computer individually with `catkin_make run_tests`. The first one didn't pass. So there may be some defects with the online test.

ipa-mdl:
I will test it again.

> So there may be some defects with the online test

Sometimes Travis tests the wrong code in case of forced pushes.
However, if the online test is what matters ;)

ipa-mdl:
I pushed the test commit to a different branch: https://travis-ci.org/ros-controls/control_toolbox/builds/482853143

src/pid.cpp
```diff
@@ -335,12 +336,11 @@ double Pid::computeCommand(double error, double error_dot, ros::Duration dt)
   // Calculate the integral of the position error
   i_error_ += dt.toSec() * p_error_;
 
-  if(gains.antiwindup_)
+  if(gains.antiwindup_ && gains.i_gain_!=0)
```
ipa-mdl: The PR is ready for merge.

There's just one smaller item to discuss:
What should happen if `i_gain_` is 0 (e.g. temporarily)?
For now `i_error_`  would accumulate..
Beforehand, the behavior was undefined.
One option would be to set the error to 0 as well.

onionsflying: That's a good question. Since `i_error_` would be clamped as soon as `i_gain_` is non-zero, I think the question is sort of like this: When `i_gain_` is changed from zero to non-zero, should `i_term` and `i_error_` be near the windup area (a. `i_error_` is accumulated when `i_gain_` is 0) or be near zero(b. `i_error_` is 0 when `i_gain_` is 0)? I think (a) is a good option too. What's your opinion?

ipa-mdl: I am okay with option a, because - as you already pointed out - the error will be clamped as soon as the gain is non-zero.
That's what would happen with `antiwindup=false` anyway.
I am just not sure if the user expects this if the gain is moved from zero to non-zero for the first time.
We could even freeze the error while the gain is zero..

onionsflying: In my opinion, the point is what the user expects when he sets `i_gain_` as zero. A PID controller whose `i_gain_` is zero (option a) or a PD controller (option b)? I agree that most users may expect option b. So how about changing the code as follows?

```diff
- i_error_ += dt.toSec() * p_error_;
+ if(gains.i_gain_==0)
+   i_error_ = 0;
+ else
+   i_error_ += dt.toSec() * p_error_;
```

ipa-mdl: For now I'd like to keep the existing behavior for `antiwindup=false`.
If you agree to reset the error to zero, please do so inside  the `gains.antiwindup_` clause:
```c++
  if(gains.antiwindup_)
  {
    if (gains.i_gain_!=0)
    {
      // Prevent i_error_ from climbing higher than permitted by i_max_/i_min_
      boost::tuple<double, double> bounds = boost::minmax<double>(gains.i_min_ / gains.i_gain_, gains.i_max_ / gains.i_gain_);
      i_error_ = boost::algorithm::clamp(i_error_, bounds.get<0>(), bounds.get<1>());
    }
    else
    {
      // Reset i_error_
      i_error_ = 0
    }
  }
```
If you prefer option a, we can merge it as-is.
@bmagyar: Do you have any preference?

bmagyar: I'd also prefer option a.

onionsflying:
Option a is good for me. Thanks for the reviewing.

PR #74, opened 2019-02-01
add erand48() implementation on windows

kejxu:
`erand48()` is not available on Windows. to work around this, add `erand48` definition using functions from STL's `random` library

this is a bit different from the behavior on Linux since this implementation with STL's functions would yield (actually) random numbers since we're not using the pre-defined [seed](https://github.com/ros-controls/control_toolbox/blob/kinetic-devel/include/control_toolbox/dither.h#L96).

However, judging from the code, we wonder if the original seed is necessary, would making these truly random cause any problem?

### DIFF
--- a/src/dither.cpp
+++ b/src/dither.cpp
@@ -36,14 +36,19 @@
 
 #include <control_toolbox/dither.h>
 
-#ifdef WIN32
+#ifdef _WIN32
 #include <random>
-#include <iostream>
-std:: default_random_engine generator;
-std::uniform_real_distribution<double> distr(0.0,1.0);
+
 double erand48(unsigned short xsubi[3])
 {
-  return distr(generator);
+  // following instructions from https://isocpp.org/files/papers/n3551.pdf
+
+  std::random_device rdev{};
+  std::default_random_engine generator{rdev()};
+
+  // uniform distribution on the interval [0.0, 1.0)
+  std::uniform_real_distribution<double> distribution(0.0, 1.0);
+  return distribution(generator);
 }
 #endif
 
--- a/src/dither.cpp
+++ b/src/dither.cpp
@@ -36,6 +36,17 @@
 
 #include <control_toolbox/dither.h>
 
+#ifdef WIN32
+#include <random>
+#include <iostream>
+std:: default_random_engine generator;
+std::uniform_real_distribution<double> distr(0.0,1.0);
+double erand48(unsigned short xsubi[3])
+{
+  return distr(generator);
+}
+#endif
+
 namespace control_toolbox {
 
 Dither::Dither() : amplitude_(0), has_saved_value_(false)


bmagyar:
I'd prefer to minimize `ifdef`s in the code as much as possible. Do you have a suggestion for replacing the current `erand48` with one that works for both platforms?

kejxu:
> I'd prefer to minimize `ifdef`s in the code as much as possible. Do you have a suggestion for replacing the current `erand48` with one that works for both platforms?

makes sense! that would be able to provide even better portability. this `ifdef` was used to minimize impact to existing Linux logic. From our understanding of [`erand48`](https://linux.die.net/man/3/erand48), it seems that
> `double erand48(unsigned short xsubi[3]);`
> The `drand48()` and `erand48()` functions return nonnegative double-precision floating-point values uniformly distributed between [0.0, 1.0).

this sounds just like what `uniform_real_distribution` from STL's `<random>` library [could provide](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution), which is the implementation added in this change. The difference (problem) is, current existing code takes in a `const double &seed` during `Dither::init`.

- if "true" randomness is wanted, we think the added `std::default_random_engine generator{rdev()};` in STL's implementation would be a great substitution here given its portability and everything
- if `seed` is still needed (question, is it?), [STL's implementation](https://en.cppreference.com/w/cpp/numeric/random) still sounds suitable since its random number generators could be initialized with certain state (seed), so it would generate the same sequence of output given the same input/seed. However, since we don't know too much about the existing code, a random number generator needs to be chosen by the maintainer =)
- `default_random_engine` is defined based on implementation, if uniform behavior is wanted, maybe a certain type of random number generator should be chosen from STL's `<random>`

ipa-mdl:
I guess, the seed was intended for predictable tests, I'd like to keep it for now.
This repo still targets kinetic, so C++11 is not an option.
Either we branch for melodic or we use `boost::random`.
@bmagyar: Are you okay with branching?

kejxu:
> I guess, the seed was intended for predictable tests, I'd like to keep it for now.

makes sense, I'll look into how seed should be fed.

> This repo still targets kinetic, so C++11 is not an option.
> Either we branch for melodic or we use `boost::random`.

just a side note, we found this in `erand48`'s [documentation](https://linux.die.net/man/3/erand48)
```
These functions are declared obsolete by SVID 3, which states that rand(3) should be used instead.
```
as for `rand()`, generally STL's `<random>` is preferred, and here's one video I found really interesting/helpful, check it out =)
https://channel9.msdn.com/Events/GoingNative/2013/rand-Considered-Harmful

kejxu:
this change is no longer necessary, closing the pull request.

meantime, hopefully we/ROS would further bring more and more c++11 into `ros-controls` in the future =)

PR #78, opened 2019-05-22
Allow smoother resets

kobytg17:
1. Currently after a reset (or init), since p_error_last is_ set to 0, the first computeCommand is calculating
a false large derivative error. Proposed fix: only calculate error_dot if p_error_last_ contains a real value.

2. Allow setting initial i_error. This allows PID to be used in a smoother manner in systems switching from
manual to automatic modes. [Reference](http://brettbeauregard.com/blog/2011/04/improving-the-beginner%E2%80%99s-pid-initialization/)

3. Improve cases where dt==0 or invalid errors were given (return last cmd_ instead of 0)

### DIFF
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -306,14 +306,16 @@ double Pid::computeCommand(double error, ros::Duration dt)
 {
 
   if (dt == ros::Duration(0.0) || std::isnan(error) || std::isinf(error))
-    return cmd_;
+    return 0.0;
 
   double error_dot = d_error_;
 
   // Calculate the derivative error
-  if (dt.toSec() > 0.0 && valid_p_error_last_)
+  if (dt.toSec() > 0.0)
   {
-    error_dot = (error - p_error_last_) / dt.toSec();
+    if (valid_p_error_last_) {
+      error_dot = (error - p_error_last_) / dt.toSec();
+    }
     p_error_last_ = error;
     valid_p_error_last_ = true;
   }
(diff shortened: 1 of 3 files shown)


src/pid.cpp
```diff
@@ -299,15 +306,16 @@ double Pid::computeCommand(double error, ros::Duration dt)
 {
 
   if (dt == ros::Duration(0.0) || std::isnan(error) || std::isinf(error))
-    return 0.0;
+    return cmd_;
```
bmagyar: What does this improve? It is unclear what a PID command should be in any of the cases in the `if` statement. The only reason we aren't throwing an exception here is to keep a simplistic, controller-friendly interface. We could change the return to `-inf` or `nan` if you wish ;)

bmagyar:
ping @ipa-mdl

src/pid.cpp
```diff
@@ -299,15 +306,16 @@ double Pid::computeCommand(double error, ros::Duration dt)
...
 
   double error_dot = d_error_;
 
   // Calculate the derivative error
-  if (dt.toSec() > 0.0)
+  if (dt.toSec() > 0.0 && valid_p_error_last_)
   {
     error_dot = (error - p_error_last_) / dt.toSec();
     p_error_last_ = error;
+    valid_p_error_last_ = true;
```
ipa-mdl: I don't understand this change. This code will only be reached if `valid_p_error_last_` is already `true`.
If it is `false`, only `reset()` can set it to `true` again.

kobytg17: My bad, I don't know how this slipped in this way.. Sorry
Added a test

kobytg17:
Can we merge this please?

kobytg17:
Do we want to merge, or should I close this PR? @ipa-mdl @bmagyar

bmagyar:
Merging for Melodic and Noetic

PR #80, opened 2019-05-23
Remove Extra include of iostream

sloretz:
`sinusoid.h` includes `<iostream>`, but it is only used in `sinusoid.cpp`. 
https://github.com/ros-controls/control_toolbox/blob/87c3ca1efedaa5740b517ac706c27d56ee6b3559/src/sinusoid.cpp#L39

This PR removes the extra include from the header since it's unused there.


bmagyar:
Good catch!

Issue #85, opened 2020-03-25
Refactor the Pid class to be completely ROS agnostic 


bmagyar:
...and create a new RosPid in a separate package that offers parameterization and services through the node interface. This would wrap the features now offered between the pid_gains_setter and the stuff coupled inside Pid::init*() functions

Split from #83

ddengster:
As I mentioned in the PR, should we put the parameter 'i_clamp' in?
https://github.com/ros-controls/control_toolbox/blob/melodic-devel/src/pid.cpp#L126

Reasoning being:
> well someone who had the i_clamp parameter and was using it in ros1 would need to substitute it for 2 more parameters when porting to ros2 - and tracking it down would not be very obvious especially when we don't generally have the time to put up changelists. Also I think we can spare the few lines of code.

matthew-reynolds:
> should we put the parameter 'i_clamp' in?

Just my 2 cents, seems redundant to me and not worth supporting for such a minor migration difference (ROS1=`i_clamp` -> ROS2=`i_clamp_min` and `i_clamp_max` is pretty painless).

ahcorde:
PR https://github.com/ros-controls/control_toolbox/issues/85

jordan-palacios:
@bmagyar As @ahcorde pointed out, I think we can close this now thanks to the work done in #97.

bmagyar:
Agreed, thanks!

PR #94, opened 2020-06-10
Added a prefix option to parameters

ahcorde:
Right now if you try to add two pid controller to the same node parameter server, the parameters are shared for the pid controllers. I added a optional argument to add a prefix to the pid controller

the original code create this parameters:
```
/gazebo_ros2_control:
  antiwindup
  d
  i
  i_clamp_max
  i_clamp_min
  p
```

This will generate something like this:

```
/gazebo_ros2_control:
  slider_to_cart.antiwindup
  slider_to_cart.d
  slider_to_cart.i
  slider_to_cart.i_clamp_max
  slider_to_cart.i_clamp_min
  slider_to_cart.p
  bar.antiwindup
  bar.d
  bar.i
  bar.i_clamp_max
  bar.i_clamp_min
  bar.p
```

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -88,19 +88,13 @@ void Pid::initPid(
   initPid(p, i, d, i_max, i_min, gains.antiwindup_, node_param_iface, parameter_prefix);
 }
 
-bool endsWith(const std::string & str, const std::string & suffix)
-{
-  return str.size() >= suffix.size() &&
-         0 == str.compare(str.size() - suffix.size(), suffix.size(), suffix);
-}
-
 void Pid::initPid(
   double p, double i, double d, double i_max, double i_min, bool antiwindup,
   NodeParamsIfacePtr node_param_iface, std::string parameter_prefix)
 
 {
   parameter_prefix_ = parameter_prefix;
-  if (!parameter_prefix_.empty() && !endsWith(parameter_prefix_, ".")) {
+  if (!parameter_prefix_.empty() && (parameter_prefix_.back() != '.')) {
     parameter_prefix_ = parameter_prefix_ + std::string(".");
   }
   initPid(p, i, d, i_max, i_min, antiwindup);
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -91,7 +91,6 @@ void Pid::initPid(
 void Pid::initPid(
   double p, double i, double d, double i_max, double i_min, bool antiwindup,
   NodeParamsIfacePtr node_param_iface, std::string parameter_prefix)
-
 {
   parameter_prefix_ = parameter_prefix;
   if (!parameter_prefix_.empty() && (parameter_prefix_.back() != '.')) {
--- a/src/pid.cpp
+++ b/src/pid.cpp
@@ -88,9 +88,10 @@ void Pid::initPid(
   initPid(p, i, d, i_max, i_min, gains.antiwindup_, node_param_iface, parameter_prefix);
 }
 
-bool endsWith(const std::string& str, const std::string& suffix)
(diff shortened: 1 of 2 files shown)


src/pid.cpp
```diff
@@ -81,16 +81,28 @@ Pid::Pid(const Pid & source)
 Pid::~Pid() {}
 
 void Pid::initPid(
-  double p, double i, double d, double i_max, double i_min, NodeParamsIfacePtr node_param_iface)
+  double p, double i, double d, double i_max, double i_min, NodeParamsIfacePtr node_param_iface,
```
Karsten1987, line 84: why is the `node_param_iface` and the `parameter_prefix` not part of the constructor?

src/pid.cpp
```diff
@@ -81,16 +81,28 @@ Pid::Pid(const Pid & source)
...
 }
 
 void Pid::initPid(
   double p, double i, double d, double i_max, double i_min, bool antiwindup,
-  NodeParamsIfacePtr node_param_iface)
+  NodeParamsIfacePtr node_param_iface, std::string parameter_prefix)
+
 {
+  parameter_prefix_ = parameter_prefix;
+  if (!parameter_prefix_.empty() && !endsWith(parameter_prefix_, ".")) {
```
Karsten1987, line 103: I don't think it needs a dedicated function for this (`endsWith`). Why not just checking for `*parameter_prefix_.end() != '.'`?

src/pid.cpp
```diff
@@ -104,12 +116,12 @@ void Pid::initPid(
...
-    declare_param("d", rclcpp::ParameterValue(d));
-    declare_param("i_clamp_max", rclcpp::ParameterValue(i_max));
-    declare_param("i_clamp_min", rclcpp::ParameterValue(i_min));
-    declare_param("antiwindup", rclcpp::ParameterValue(antiwindup));
+    declare_param(parameter_prefix_ + "p", rclcpp::ParameterValue(p));
+    declare_param(parameter_prefix_ + "i", rclcpp::ParameterValue(i));
+    declare_param(parameter_prefix_ + "d", rclcpp::ParameterValue(d));
+    declare_param(parameter_prefix_ + "i_clamp_max", rclcpp::ParameterValue(i_max));
+    declare_param(parameter_prefix_ + "i_clamp_min", rclcpp::ParameterValue(i_min));
+    declare_param(parameter_prefix_ + "antiwindup", rclcpp::ParameterValue(antiwindup));
```
Karsten1987, line 124: I don't think that the `declare_param` calls should be part of the init function, but rather be declared within the ctor.

bmagyar:
Related to this PR but may be out of scope: when @jordan-palacios worked on this before he highlighted that it'd be much cleaner if we could separate the node-specific items from the PID-specifics. Having a pure C++ PID class *and* a ROS wrapper around it would be a much cleaner way to go.

src/pid.cpp
ahcorde: Fixed https://github.com/ros-controls/control_toolbox/pull/94/commits/14f52e3bca5180c19a2a653433d742b0516adc8c

ahcorde:
@Karsten1987, I think this is because this implementation try to split the PID from ROS. There is an open ticket for this https://github.com/ros-controls/control_toolbox/issues/85

Karsten1987:
so if the point is to have a ROS agnostic PID class (which I fully support btw) then I don't think this PR is the right fix as it's even introducing more ROS specifics.

ahcorde:
Closing this PR in favour of this other one https://github.com/ros-controls/control_toolbox/pull/97

PR #109, opened 2021-06-12
Add Control Filters - Gravity Compensation

destogl:
This PR ports the filters used in controllers initially implemented in: https://github.com/KITrobotics/iirob_filters.

Reason for this is dependency of FTS Broadcaster and AdmittanceController on those and to make this useful filter more accessible.

The filters leverage ros/filters as base enabling dynamic loading and creating filter chaing using configurations in yaml files.


destogl:
Closed in favor of #115

PR #113, opened 2021-07-13
Increased integral gains & windup limits from +-1000 to +-100000

stevegolton:
In a recent application we have found that we needed an a integral gain of around 7500 - 10000, which is rather larger than the +-1000 limits enforced by the dynamic reconfigure parameters in `Parameters.cfg`. Whether having an integral gain this high is actually a good idea or not is another matter... but it would be nice to be able to try it out without being constrained by artificial limits :smiley:

This change simply increases the limits of the integral gain from +-1000 to +-100000 which provides ample wiggle room, at the expense of the sliders being harder to fine tune in dynamic reconfigure. Seeing as the P gain limits are already +-100000, I figured this wouldn't be too much of an issue.


destogl (approved):
This looks OK and reasonable to me

bmagyar (approved):

PR #116, opened 2021-10-30
Add Parameter Handler to simplify management of params in filters and controllers and simply enable dynamic reconfigure

destogl:
- Add Parameter Handler class.
- Add support for all available parameter types.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,10 +20,14 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(control_msgs REQUIRED)
+find_package(rcl_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(realtime_tools REQUIRED)
 
+########################
+# Control toolbox
+########################
 add_library(${PROJECT_NAME} SHARED
   src/dither.cpp
   src/limited_proxy.cpp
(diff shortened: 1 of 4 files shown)


bmagyar:
FYI @tylerjw

tylerjw (commented):
My problem with this as a generic interface is it uses the ros2 interfaces directly for declaring and getting parameters.  I think this would be much more useful (and future prof) if it didn't abstract away the interface for declaring (and getting) parameters.  One of the really awesome parts of ros2's parameter system is that parameters can describe themselves, validate basic properties like ranges, and ros2 can dump the default config from any node.

I don't think any pattern or tool for using parameters in ROS2 should abstract away the interface in ROS2 itself and should instead just provide structure for the various steps a user needs to implement (declare, get, dynamic update) in a straightforward way.

destogl:
Hi @tylerjw,

Thanks for the review.

> My problem with this as a generic interface is it uses the ros2 interfaces directly for declaring and getting parameters. I think this would be much more useful (and future prof) if it didn't abstract away the interface for declaring (and getting) parameters. 

This is a very valid comment. You are right! I was also disturbed with something, but in my “bubble” couldn't see with what.

> I don't think any pattern or tool for using parameters in ROS2 should abstract away the interface in ROS2 itself and should instead just provide structure for the various steps a user needs to implement (declare, get, dynamic update) in a straightforward way.

I will think about this over the weekend and check how this can be achieved and simplified.

destogl:
Note to myself:

check also this out https://github.com/ros2/rclcpp/pull/1796 (thanks @tylerjw)

destogl:
This is obsolete now! Use "generate_parameters_library" repository.

PR #117, opened 2022-01-26
Switch to std::bind

jspricke:


destogl (approved):
Looking good to me!

bmagyar (changes requested):
actually... not for melodic, sorry

let me create a master branch or something else wehre new stuff can go to

jspricke:
std::bind is part of C++11 so should be fine in melodic, right? https://en.cppreference.com/w/cpp/utility/functional/bind

bmagyar (approved):

Issue #118, opened 2022-02-22
PID controller handling of parameters in ROS 2


SammyRamone:
I was switching my code, which uses the PID controller from this project, from ROS1 to ROS2 and ran into the following issue:
When the PID controller is created a reference on the node is taken as a parameter. I assumed that this should be a reference to the node that is using this PID controller. The description of the constructor (https://github.com/ros-controls/control_toolbox/blob/ros2-master/include/control_toolbox/pid_ros.hpp#L58-L61) seemed to suggest that I only need to put it in a correct namespace.
Therefore, I programmed something like this:
```
std::shared_ptr<rclcpp::Node> node = rclcpp::Node::make_shared("node");
node->declare_parameter<double>("pid.p", 0.0);
node->declare_parameter<double>("pid.i", 0.0);
node->declare_parameter<double>("pid.d", 0.0);
node->declare_parameter<double>("pid.i_clamp_max", 0.0);
node->declare_parameter<double>("pid.i_clamp_min", 0.0);
node->declare_parameter<bool>("pid.antiwindup", false);
control_toolbox::PidROS pid = control_toolbox::PidROS(node, "pid");
pid.initPid();
node->declare_parameter<double>("not_related_to_pid", 0.0);
double not_related_to_pid;
node->get_parameter("not_related_to_pid", not_related_to_pid);
```
Executing this leads to following error:
```
[Node-4] terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterValueException'
[Node-4]   what():  parameter 'not_related_to_pid' could not be set: Invalid parameter

```
I think, the reason is that the pid controller sets the `add_on_set_parameters_callback` here: https://github.com/ros-controls/control_toolbox/blob/d7462e982e40bfdebddbcff8849c611caf8c8422/src/pid_ros.cpp#L281-L326
This is now for the whole node and therefore any other not PID related parameters will result in errors.

I managed to fix this by creating a seperate node for the pid controller.
```
std::shared_ptr<rclcpp::Node> node = rclcpp::Node::make_shared("node");
std::shared_ptr<rclcpp::Node> node_pid = rclcpp::Node::make_shared("pid");
node_pid->declare_parameter<double>("p", 0.0);
node_pid->declare_parameter<double>("i", 0.0);
node_pid->declare_parameter<double>("d", 0.0);
node_pid->declare_parameter<double>("i_clamp_max", 0.0);
node_pid->declare_parameter<double>("i_clamp_min", 0.0);
[...]

PR #119, opened 2022-03-11
[PidROS] Add topic_prefix_ to declareParam & setParameter

chenjunnn:
- Fix the issue when setting the parameters of `PidROS` with topic prefix
```shell
# if prefix = "pitch"
$ ros2 param set /gimbal_controller pitch.p 1.5
Setting parameter failed: Invalid parameter
```
- Add getParametersCallbackHandle function so that user can remove set_parameters_callback to avoid issue like https://github.com/ros-controls/control_toolbox/issues/118
```cpp
node_->remove_on_set_parameters_callback(pid_ros_->getParametersCallbackHandle().get());
```

### DIFF
--- a/include/control_toolbox/pid_ros.hpp
+++ b/include/control_toolbox/pid_ros.hpp
@@ -177,6 +177,16 @@ public:
    */
   void printValues();
 
+  /*!
+   * \brief Return PID parameters callback handle
+   * \return shared_ptr to the PID parameters callback handle
+   */
+  inline rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr
+  getParametersCallbackHandle()
+  {
+    return parameter_callback_;
+  }
+
 private:
   void setParameterEventCallback();
 
--- a/src/pid_ros.cpp
+++ b/src/pid_ros.cpp
@@ -141,12 +141,12 @@ PidROS::initPid(double p, double i, double d, double i_max, double i_min, bool a
 {
   pid_.initPid(p, i, d, i_max, i_min, antiwindup);
 
-  declareParam("p", rclcpp::ParameterValue(p));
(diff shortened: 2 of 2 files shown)


include/control_toolbox/pid_ros.hpp
```diff
@@ -122,6 +122,19 @@ class CONTROL_TOOLBOX_PUBLIC PidROS
...
+   * time step size. This also allows the user to pass in a precomputed
+   * derivative error.
+   *
+   * \param error Error since last call (error = target - state)
+   * \param error_dot d(Error)/dt since last call
+   * \param dt Change in time since last call in seconds
+   *
+   * \returns PID command
+   */
+  double computeCommand(double error, double error_dot, rclcpp::Duration dt);
```
destogl, line 136: Do we need additional method for this? Why not always calculate `error_dot` in the `computerCommand` method and use only one of them? I understand there are two underlaying methods, but why?

Can you give more information about this change?

Also, I propose to add this change separately into another PR because it is unrelated to the PR title and paramter feature.

destogl (changes requested):
Thanks for submitting the PR. I like what you did there. Please consider my comment regarding addition of `computeCommand` method.

include/control_toolbox/pid_ros.hpp
chenjunnn: Because I'm trying to used IMU data to do the PID control. And I'm using calculated angle from IMU to compute the `error` and using angular velocity directly read from IMU as the `error_dot`.

I will open another PR to commit this change. Thank you for your reply.

destogl (approved):
Generally agree with this changes!

I would open as followup an issue about renaming `topic_prefix_` because its name is misleading, `parameter_perfix_` would be more suitable.

PR #125, opened 2022-06-30
Fix namespace collision and parameter_callback problems in PidROS

progtologist:
This fixes problems in PidROS with using the parameters (reconfiguration fails when the node has extra parameters declared) and with namespacing issues (topic namespace argument can lead to problems if defined deeper than one)

### DIFF
--- a/src/pid_ros.cpp
+++ b/src/pid_ros.cpp
@@ -44,15 +44,33 @@ namespace control_toolbox
 
 void PidROS::initialize(std::string topic_prefix)
 {
-  if (topic_prefix.back() != '.' && !topic_prefix.empty()) {
-    topic_prefix_ = topic_prefix + ".";
-  } else {
-    topic_prefix_ = topic_prefix;
+  param_prefix_ = topic_prefix;
+  // If it starts with a "~", remove it
+  if (param_prefix_.compare(0, 1, "~") == 0) {
+    param_prefix_.erase(0, 1);
+  }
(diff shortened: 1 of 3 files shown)


src/pid_ros.cpp
```diff
@@ -294,40 +312,45 @@ PidROS::setParameterEventCallback()
...
+          } else if (param_name == param_prefix_ + "i_clamp_min") {
             gains.i_min_ = parameter.get_value<double>();
-          } else if (param_name == topic_prefix_ + "antiwindup") {
+            changed = true;
+          } else if (param_name == param_prefix_ + "antiwindup") {
             gains.antiwindup_ = parameter.get_value<bool>();
-          } else {
-            result.successful = false;
-            result.reason = "Invalid parameter";
+            changed = true;
```
destogl, line 337: Why this change?

destogl (approved):
Very nice solution without API breaks! Thanks!

src/pid_ros.cpp
progtologist: The node_params_->add_on_set_parameters_callback() adds a callback that will be called whenever any parameter of the node changes. E.g. if you have 2 instances of PidROS, or if you have your own parameters in the same node, then this callback will be called even if a parameter not declared by PidROS changes. If we set result.successful to false because the parameter is not known to PidROS, we are setting false the reconfiguration of the node in general.

PR #130, opened 2023-01-10
🔧Fixes and updated on pre-commit hooks and their action.

destogl:
- Use pydocstyle instead of pep257 because it replaces it.
- Bump versions of hooks.
- Remove fixed python version from formatting action.
- Fix action warnings about Node.js 12 by updating them.
- Bump version of clang-format.

#131 corrects formatting - not sure why clang-format is passing here and not locally...


bmagyar:
I'm sooooryyy I didn't see this.o e! Could you please rebase?

PR #132, opened 2023-01-13
Update pid documentation to match src code

huemerj:
I updated the documentation for the PID-control class in include/pid.hpp to match the actual behavior from src/pid.cpp.
In PID-control literature, the error is typically defined as 'error=target-state' and the computeCommand function is also used in this way (e.g. https://github.com/ros-controls/ros2_controllers/blob/a49a87da1df40851a80507d61931d12c969cb015/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L245).

Related to #39

### DIFF
--- a/include/control_toolbox/pid.hpp
+++ b/include/control_toolbox/pid.hpp
@@ -60,7 +60,7 @@ namespace control_toolbox
   In particular, this class implements the standard
   pid equation:
 
-  \f$command  = -p_{term} - i_{term} - d_{term} \f$
+  \f$command  = p_{term} + i_{term} + d_{term} \f$
 
   where: <br>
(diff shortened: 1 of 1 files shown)


bmagyar:
I've only spotted this PR after I merged #138 

Could you please check if the current state is satisfsctory or this PR still has changes to add?

christophfroehlich:
damn, sorry. Miscommunication between @huemerj and myself. Credits go to him for finding it ;)

huemerj:
> I've only spotted this PR after I merged #138
> 
> Could you please check if the current state is satisfsctory or this PR still has changes to add?

Current state is totally fine, thank you 👍

PR #139, opened 2023-01-15
Check for i_min <= i_max at initialization

### LINKED ISSUE #102: Check that i_min is less than i_max
Using this toolbox, I was confused by the order of arguments in the constructor:

```
    Gains(double p, double i, double d, double i_max, double i_min)
       : p_gain_(p),
         i_gain_(i),
         d_gain_(d),
         i_max_(i_max),
         i_min_(i_min),
         antiwindup_(false)
     {}
```

It would be nice to throw an error if `i_min` >= `i_max`

christophfroehlich:
Closes #102


test/pid_tests.cpp
```diff
@@ -43,23 +43,11 @@ TEST(ParameterTest, ITermBadIBoundsTest)
 {
   RecordProperty(
     "description",
-    "This test checks that the integral contribution is robust to bad i_bounds specification (i.e. "
+    "This test checks that the constructor throws an error for bad i_bounds specification (i.e. "
```
bmagyar, line 46: Well technically it's "setGains" that throws. Can we trigger a throw with dyn reconf?

christophfroehlich: never used dynamic reconfigure up to now, I'll see if I can write a test for that.

christophfroehlich: @bmagyar I checked now how dynamic reconf works with the ROS-PID. So you mean if it is a good idea to throw an error in this case? I'm not sure how this is used in projects.. we could just give a RCLCPP_WARN message and forget the new parameters, but throw an error only if the constructor is called with i_min>i_max?

src/pid.cpp
```diff
@@ -126,6 +126,9 @@ void Pid::setGains(double p, double i, double d, double i_max, double i_min, boo
 
 void Pid::setGains(const Gains & gains)
 {
+  if (gains.i_min_ > gains.i_max_) {
```
destogl, line 129: Should we also check this in the line 120?

src/pid.cpp
```diff
@@ -126,6 +126,9 @@ void Pid::setGains(double p, double i, double d, double i_max, double i_min, boo
...
+    throw std::invalid_argument("received i_min > i_max");
```
destogl, line 130: IMHO, it would be better to add ERROR here and disable PID then throw exception. What if we have running system and start new controller with bad parameter – then the whole drive crashes. If we disable movement of something would be better, wouldn't be?

Or should we catch this in controller manager, to prevent HW drivers and other controllers to crash?

destogl (commented):
We should think about a whole use-case / concept to handle this. We don't have to solve it, but use the situation to think...

src/pid.cpp
christophfroehlich: we could, but in the end `void Pid::setGains(const Gains & gains)` is called anyways.

src/pid.cpp
christophfroehlich: with ERROR, you mean an output on `std::cerr`?

"Disable movement" is not straight forward, isn't it? It could be in some configuration where zero torque would cause the robot to crash -> we cannot solve all failure cases from this generic controller.

I think for the initial setup of the controllers it is safe to "just" throw an exception (nothing is moving, we are in a safe configuration)? In the case of online reconfiguration, we also could just drop the new gains, give an error message, and work with the previous ones.

test/pid_tests.cpp
destogl: @bmagyar I think we have to this in the controller. here we can only document the behavior

src/pid.cpp
christophfroehlich: I changed it now, so that the exception is only thrown if the base class is constructed with bad parameters. Otherwise, the new gains are skipped and nothing more happens.

bmagyar (dismissed):
Thanks!

bmagyar (approved):

src/pid_ros.cpp
```diff
@@ -307,7 +315,14 @@ void PidROS::printValues()
                                                     << "  Command:      " << getCurrentCmd(););
 }
 
-void PidROS::setGains(const Pid::Gains & gains) { pid_.setGains(gains); }
+void PidROS::setGains(const Pid::Gains & gains)
+{
+  if (gains.i_min_ > gains.i_max_) {
+    RCLCPP_ERROR(node_logging_->get_logger(), "received i_min > i_max, skip new gains");
+  } else {
+  pid_.setGains(gains);
```
destogl, line 323:
```suggestion
    pid_.setGains(gains);
```

src/pid_ros.cpp
```diff
@@ -348,7 +363,11 @@ void PidROS::setParameterEventCallback()
 
     if (changed) {
       /// @note don't call setGains() from inside a callback
-      pid_.setGains(gains);
+      if (gains.i_min_ > gains.i_max_) {
+        RCLCPP_ERROR(node_logging_->get_logger(), "received i_min > i_max, skip new gains");
+      } else {
+          pid_.setGains(gains);
```
destogl, line 369:
```suggestion
        pid_.setGains(gains);
```

destogl (approved):
Thanks

PR #140, opened 2023-01-15
Use std::clamp

christophfroehlich:
should work with C++17 since galactic


destogl:
@bmagyar we have to branch out for foxy now...

destogl (approved):
Thanks!

bmagyar (approved):
thxxx

PR #148, opened 2023-02-21
Enable subclassing of PID implementation.

destogl:
In some cases, internal values of PID have to be manipulated to work on hardware. This change enables to subclass the PID implementation and adjust internal states as needed.

### DIFF
--- a/include/control_toolbox/pid.hpp
+++ b/include/control_toolbox/pid.hpp
@@ -279,7 +279,7 @@ public:
     return *this;
   }
 
-private:
+protected:
   // Store the PID gains in a realtime buffer to allow dynamic reconfigure to update it without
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):

christophfroehlich:
I'm just curious, which states do you have to manipulate (except for the gains)?

bijoua29 (approved):

bijoua29:
> I'm just curious, which states do you have to manipulate (except for the gains)?

Mainly used to subclass for some alternative implementations. For example, we subclass to implement an adaptive-based PID.

bmagyar (approved):
Thanks! I presume you only target rolling with this?

PR #115, opened 2021-09-23
Ported LowPass and GravityCompensation filters iirob_filters repository

destogl:
- using the dynamic parametrization as implemented in existing gravity_compensation filter
- added testcases for loading parameters

PR #152, opened 2023-03-07
Add lowpass filter

guihomework:
This is a rework of PR #115 for the low-pass filter part, now using `generate_parameter_library`.


guihomework:
xmlcheck complained about angle brackets, to type of class to instantiate.
since there is no cli for pluginlib in ros2 yet, to quickly test if the <> would be usable in the command line, a pluginlib instantiation should be tested within the tests

guihomework:
Test passes with the escaped angle brackets.

destogl:
@bmagyar can we branch-out here for humble and rolling?, i.e., we need one `foxy` branch that holds current code.

CMakeLists.txt
```diff
@@ -41,6 +41,45 @@ target_include_directories(control_toolbox PUBLIC
...
+
+########################
+# Build control filters
+########################
+find_package(filters REQUIRED)
+find_package(generate_parameter_library REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(Eigen3 REQUIRED)
```
destogl, line 53: Please add this first in the `CONTROL_FILTERS_INCLUDE_DEPENDS` list and then iterate over the list ot include all dependencies. This reduces duplication, and it makes easier maintenance of the file.

CMakeLists.txt
```diff
@@ -41,6 +41,45 @@ target_include_directories(control_toolbox PUBLIC
...
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<BUILD_INTERFACE:${EIGEN3_INCLUDE_DIR}>
+    $<INSTALL_INTERFACE:include>
+)
+generate_parameter_library(
+  low_pass_filter_parameters
+  src/control_filters/low_pass_filter_parameters.yaml
+)
+target_link_libraries(low_pass_filter low_pass_filter_parameters)
+ament_target_dependencies(low_pass_filter ${CONTROL_FILTERS_INCLUDE_DEPENDS})
```
destogl, line 78:
```suggestion
target_compile_features(low_pass_filter PUBLIC cxx_std_17)
target_include_directories(low_pass_filter PUBLIC
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
    $<BUILD_INTERFACE:${EIGEN3_INCLUDE_DIR}>
    $<INSTALL_INTERFACE:include/control_filters>
)
target_link_libraries(low_pass_filter PUBLIC low_pass_filter_parameters)
ament_target_dependencies(low_pass_filter ${CONTROL_FILTERS_INCLUDE_DEPENDS})
```

CMakeLists.txt
```diff
@@ -41,6 +41,45 @@ target_include_directories(control_toolbox PUBLIC
...
+
+set(CONTROL_FILTERS_INCLUDE_DEPENDS
+  Eigen3
+  generate_parameter_library
+  filters
+  geometry_msgs
+  pluginlib
+  rclcpp
+)
+
```
destogl, line 63:
```suggestion

generate_parameter_library(low_pass_filter_parameters
  src/control_filters/low_pass_filter_parameters.yaml
)

```

CMakeLists.txt
destogl: This line `$<INSTALL_INTERFACE:include/control_filters>`  maybe have to be `$<INSTALL_INTERFACE:include/control_toolbox>` - please test which version is working.

CMakeLists.txt
```diff
@@ -55,19 +94,37 @@ if(BUILD_TESTING)
...
-install(TARGETS control_toolbox
+install(TARGETS control_toolbox low_pass_filter low_pass_filter_parameters
   EXPORT export_control_toolbox
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
 
+# Install pluginlib xml
+pluginlib_export_plugin_description_file(filters control_filters.xml)
```
destogl, line 121: We add this part usually before tests.

CMakeLists.txt
```diff
@@ -55,19 +94,37 @@ if(BUILD_TESTING)
...
+
 ament_export_targets(export_control_toolbox HAS_LIBRARY_TARGET)
-ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
+ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS} ${CONTROL_FILTERS_INCLUDE_DEPENDS})
+ament_export_libraries(
+  control_toolbox
+  low_pass_filter
+  low_pass_filter_parameters
+)
```
destogl, line 129: This should not be necessary because we are exporting targets 2 lines above as they are defined in line 113-114.

README.md
```diff
@@ -27,3 +27,25 @@ doi = {10.21105/joss.00456},
...
+To run it initially over the whole repo you can use:
+  ```
+  pre-commit run -a
+  ```
+
+If you get error that something is missing on your computer, do the following for:
+
+  - `clang-format-10`
+     ```
+     sudo apt install clang-format-10
```
destogl, line 50: We use clang-format-14
```suggestion
  - `clang-format-14`
     ```
     sudo apt install clang-format-14
```

See here: https://github.com/ros-controls/control_toolbox/blob/ros2-master/.github/workflows/ci-format.yml

control_filters.xml
```diff
@@ -0,0 +1,18 @@
+<class_libraries>
+    <library path="low_pass_filter">
+      <class name="control_filters/LowPassFilterDouble"
+             type="control_filters::LowPassFilter&lt;double&gt;"
+             base_class_type="filters::FilterBase&lt;double&gt;">
```
destogl, line 5: This nomenclature doesn't look nice... It hinders readability very much. We could ignore this field in `xml` checks. But this is also not great.

@bmagyar which "evil" is less?

include/control_filters/low_pass_filter.hpp
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) 2021, Stogl Robotics Consulting UG (haftungsbeschränkt)
```
destogl, line 1: Let's bump the year.

```suggestion
// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt)
```

destogl (commented):
1st part...

control_filters.xml
guihomework: The syntax is also used in the low-level base package https://github.com/ros/filters/blob/ros2/default_plugins.xml

CMakeLists.txt
guihomework: i am actually surprised one needs to install the includes in a different folder than the project name or our own name (control_filters).
The code before my PR #152 (which is only 3 weeks old from #145 ) already installs in a control_toolbox sub-folder of the project name, making it look like 
`my_ws/install/control_toolbox/include/control_toolbox/control_toolbox`
`my_ws/install/control_toolbox/include/control_toolbox/control_filters`
is that really what we want ? I have difficulties understanding the "override" thing, so maybe there is a good reason, otherwise I would expect:
`my_ws/install/control_toolbox/include/control_toolbox`
and for the control filters
`my_ws/install/control_toolbox/include/control_filters`

it is possible this got un-noticed, because with underlays still having the control_toolbox includes looking as I would expect it, this simply worked (the underlay libs were found ?) until this new change gets released ?

destogl: I understand this is confusing. It is also for all of us, but this is how ament und colcon work...

you can read here more details and check references why we have to do that: https://github.com/ros-controls/ros2_control/pull/926

destogl: > it is possible this got un-noticed, because with underlays still having the control_toolbox includes looking as I would expect it, this simply worked (the underlay libs were found ?) until this new change gets released ?

You are adding here new libs, and they work without problem. The issue comes when you have a lib, install in the system and want to override it as in ROS 1 – this doesn't work without these tricks. After loosing already a dozen of hours debugging non-existing issues, we decided to take this ugly implementation and just make things to be easily used.

control_filters.xml
destogl: @guihomework a good point! Then we should go this way. @bmagyar do you agree?

README.md
destogl: Please let the reviewer close/mark the comments as resolved. You can just give a short remark like "done" or thumbs up emoji if you just agree. Sometimes is confusing when something is marked as done, but actually one doesn't sees it in the changes.

CMakeLists.txt
guihomework: OK, maybe #145 should have at least pointed to the doc justifying that, then I would not have asked.

This is not nice but very important and something others will stumble upon. I had seen the override warnings in colcon build but I did not relate the "ordering" to the strange naming.

I see that ROS 1 solved that known issue of overlays by having re-ordering of the includes done in the cmake when catkin was called.

I hope this gets solved one day.

README.md
guihomework: Oh ok, this is how it works. I did not use it this way in the past.

CMakeLists.txt
guihomework: > This line `$<INSTALL_INTERFACE:include/control_filters>` maybe have to be `$<INSTALL_INTERFACE:include/control_toolbox>` - please test which version is working.

since the destination of the source includes installation is this name-trick to permit override, the install interface of the filters has to be in the same folder as this name-trick. 
if one wants to have a different name-trick applied to the `control_filters` so that it goes in `control_filters/control_filters`
the install cmake command must be modified to do 2 sub-folder installs for `control_toolbox` and `control_filters` separately.

depends what you prefer:

`include/control_toolbox/control_toolbox`
`include/control_toolbox/control_filters`
or
`include/control_toolbox/control_toolbox`
`include/control_filters/control_filters`

guihomework: For my understanding : 
Could you confirm there is no override issue with the generated libraries headers (in case a this generate library yaml gets changed in an overlay), although they get installed in `include/low_pass_filter_parameters/` and not in `include/low_pass_filter_parameters/low_pass_filter_parameters` because the `target_link_library` sees it is a local target and finds the local header  ?

test/control_filters/test_low_pass_filter.cpp
```diff
@@ -0,0 +1,120 @@
...
+    ASSERT_TRUE(filter_->configure("", "TestLowPassFilter",
+        node_->get_node_logging_interface(), node_->get_node_parameters_interface()));
+}
+
+TEST_F(LowPassFilterTest, TestLowPassFilter)
+{
+    double sampling_freq = 1000.0;
+    double damping_freq = 20.5;
+    double damping_intensity = 1.25;
+    int divider = 3;
```
christophfroehlich, line 52: same as above

christophfroehlich (changes requested):

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,37 @@
...
+  damping_intensity: {
+    type: double,
+    # default_value: 1.25,
+    description: "Damping intensity",
+    #validation: {
+    #  gt<>: [0.0]
+    #},
+  }
+
+  #divider:
```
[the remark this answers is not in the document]
guihomework: I kept it in this draft to question @destogl on what this parameter was supposed to be (was not used in the previous PR) but forgot to put the actual question here. Thanks for asking. @destogl, what was this _divider_ meant to be ?

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,37 @@
...
+    description: "Sampling frequency",
+    validation: {
+      gt<>: [0.0]
+    },
+  }
+  damping_frequency: {
+    type: double,
+    # default_value: 20.5,
+    description: "Damping frequency",
+    #validation: {
```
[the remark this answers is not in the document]
guihomework: I was not 100% sure if imposing a condition was necessary, if I am not mistaken, the filter equation would not fail with any values, so enforcing is probably just for common sense, not for protection against mathematical impossibilities (like DIV/0 for the _sampling_frequency_)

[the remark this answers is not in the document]
christophfroehlich: The filter becomes unstable with negative damping frequency. I don't know why this should be allowed here
image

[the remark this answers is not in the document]
guihomework: thanks a lot for running a test. I am not yet totally configured on my 22.04 recent installation, and did not yet prepare all the tools I would usually have for testing such things ( I would probably have put the equation in octave and prepare plots there ).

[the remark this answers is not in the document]
christophfroehlich: I had Matlab opened already ;)

include/control_filters/low_pass_filter.hpp
```diff
@@ -0,0 +1,182 @@
...
+  logger_.reset(
+    new rclcpp::Logger(this->logging_interface_->get_logger().get_child(this->filter_name_)));
+
+  // Initialize the parameters once
+  if (!parameter_handler_)
+  {
+    try
+    {
+      parameter_handler_ =
+        std::make_shared<low_pass_filter::ParamListener>(this->params_interface_, this->param_prefix_);
```
guihomework, line 90: this->param_prefix_ was missing but was not detected by the current tests because one just loads and instantiates the filter with no configure, and the second test configures with local parameters.
I suppose a rostest with param loading from yaml is necessary to ensure loading the filters is feasible within a namespace.

control_filters.xml
bmagyar: Feel like you are using Internet Explorer in the past :D nah guys it's ok, it's mostly readable

CMakeLists.txt
destogl: We didn't test this yet with generated parameters.

README.md
```diff
@@ -27,3 +27,25 @@ doi = {10.21105/joss.00456},
 URL = {http://www.theoj.org/joss-papers/joss.00456/10.21105.joss.00456.pdf}
 }
 ```
+
+
+## Code Formatting
```
destogl, line 32: We should add this in the separate PR. It's not part of the functionality presented here

include/control_filters/low_pass_filter.hpp
```diff
@@ -0,0 +1,186 @@
...
+  };
+
+private:
+  rclcpp::Clock::SharedPtr clock_;
+  std::shared_ptr<rclcpp::Logger> logger_;
+  std::shared_ptr<low_pass_filter::ParamListener> parameter_handler_;
+  low_pass_filter::Params parameters_;
+
+  // Filter parameters
+  // TODO(destogl): we should do this more intelligently using only one set of types
```
destogl, line 60: Not necessary since we didn't come up with any better solution.

```suggestion
```

src/control_filters/low_pass_filter_parameters.yaml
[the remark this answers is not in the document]
destogl: Nothing anymore. We can remove this.

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,34 @@
...
+  #{
+  #  type: int,
+  #  # default_value: 3,
+  #  description: "Divider",
+  #  validation:
+  #  {
+  #    gt_eq<>: [0]
+  #  },
+  #  read_only: true
+  #}
```
destogl, line 34:
```suggestion
```

test/control_filters/test_low_pass_filter.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "test_low_pass_filter.hpp"
+
+TEST_F(LowPassFilterTest, TestLowPassFilterParameters)
+{
+    std::shared_ptr<filters::FilterBase<geometry_msgs::msg::WrenchStamped>> filter_ =
+        std::make_shared<control_filters::LowPassFilter<geometry_msgs::msg::WrenchStamped>>();
+
+    node_->declare_parameter("damping_frequency", 20.5);
+    node_->declare_parameter("damping_intensity", 1.25);
+    node_->declare_parameter("divider", 3);
```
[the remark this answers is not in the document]
destogl: Yes, let's delete this.

```suggestion
```

test/control_filters/test_low_pass_filter.cpp
destogl:
```suggestion
```

destogl (changes requested):
Except the small issues, I am happy with this.

include/control_filters/low_pass_filter.hpp
guihomework: This was not worked on yet. I had a quick look, maybe something along using std::enable_if to detect if a type can be directly filtered with the y(n)= b * x(n-1) + a * y(n-1) example double, ints Eigen<N,1>, etc..) or if a specialization has been provided for converting the T type to a processable storage...
https://medium.com/@sidbhasin82/c-templates-what-is-std-enable-if-and-how-to-use-it-fd76d3abbabe

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,22 @@
+low_pass_filter:
+  sampling_frequency: {
+    type: double,
+    # default_value: 1000.0,
```
destogl, line 4:
```suggestion
```

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,22 @@
...
+    description: "Sampling frequency",
+    validation: {
+      gt<>: [0.0]
+    },
+  }
+  damping_frequency: {
+    type: double,
+    # default_value: 20.5,
```
destogl, line 12:
```suggestion
```

src/control_filters/low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,22 @@
...
+    description: "Damping frequency",
+    validation: {
+      gt<>: [0.0]
+    },
+  }
+  damping_intensity: {
+    type: double,
+    # default_value: 1.25,
```
destogl, line 20:
```suggestion
```

test/control_filters/test_low_pass_filter.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+TEST_F(LowPassFilterTest, TestLowPassWrenchFilterMissingParameter)
+{
+    std::shared_ptr<filters::FilterBase<geometry_msgs::msg::WrenchStamped>> filter_ =
+        std::make_shared<control_filters::LowPassFilter<geometry_msgs::msg::WrenchStamped>>();
+
+    // should deny configuration as sampling frequency is missing
+    ASSERT_FALSE(filter_->configure("", "TestLowPassFilter",
+        node_->get_node_logging_interface(), node_->get_node_parameters_interface()));
+}
```
destogl, line 42: not needed when using generate_parameter_library

test/control_filters/test_low_pass_filter_parameters.yaml
```diff
@@ -0,0 +1,22 @@
...
+  ros__parameters:
+    sampling_frequency: 1000.0
+    damping_frequency: 20.5
+    damping_intensity: 1.25
+
+TestLowPassWrenchFilterMissingParameter:
+  ros__parameters:
+    damping_frequency: 20.5
+    damping_intensity: 1.25
+
```
destogl, line 11: Not needed since using generate_parameter library.

destogl (dismissed):
This looks great! Thanks. Just a few more cleanings. One test is duplicate since using generate_paramter_library and since it it done in a test later.

test/control_filters/test_low_pass_filter.cpp
guihomework: I don't see why one should not test for `filter configure` failing "properly". This test permits to verify that the parameters were correctly defined in the `generate_parameter_library` YAML as "should fail if missing", and that the throw is caught correctly isn't it ?

src/control_filters/low_pass_filter_parameters.yaml
guihomework: done all of these 3 default removal

christophfroehlich (changes requested):
If I were allowed to wish something: a short README would help guiding new users, as well as adding some function API documentation inside the header (like in all other functions of this repository).

guihomework:
> If I were allowed to wish something: a short README would help guiding new users

Do you mean a doc similar to the one in the Gravity Compensation "filter" https://github.com/ros-controls/control_toolbox/blob/f181fe76279860ff9b1c9f3d9edde927da29703b/src/control_filters/README.md
or more a general description of the role of control filters and why they are there and how to use them ?

guihomework:
@christophfroehlich  Thanks for the clarification.  Another question: I am new to PR that require rebasing+changes, when the reviewer does merges of master in the PR. What should I do on my side ? rebase my initial branch (before your merge from master) on top of maaster, commit the requested changes and push or now add the requested changes over your merged-commit. I am used to always keep history of branches clean and straight.

christophfroehlich:
I also prefer having a clean history, but in the end everything gets squashed in a single commit on master -> I'd suggest that you just commit on top of the last merge-commit.

guihomework:
This makes sense. When I kept branches clean, it was to keep all commits in the master as well.

guihomework:
> If I were allowed to wish something: a short README would help guiding new users, as well as adding some function API documentation inside the header (like in all other functions of this repository).

both done now.

christophfroehlich (approved):
Great, thx!

bmagyar (approved):

PR #153, opened 2023-03-07
Add gravity compensation filter

guihomework:
This is a rework of PR https://github.com/ros-controls/control_toolbox/pull/115 for the gravity-compensation filter part, now using generate_parameter_library.


guihomework:
The new changes to this PR stem from using the filter in admittance control and realizing mistakes.

One major mistake was the switch to wrench **transforms** (as suggested in a TODO in previous version of this code), because it (correctly) applies an **additional** torque, which should not happen due to gravity being a **field** and not a force exerted at a certain point than generates a torque when moved. So one should not transform the wrench to base or what ever other computation frame, just **rotate** it, and remove the gravity force on the correct axes.

Additionally, since there are `lookupTransforms` done, I offered the possibility to have a different output frame than the input frame. A transform back was pre-existing anyway, so might as well be one that is useful if desired.

I refactored variable names and comments to better explain what happens, and simplified the frames (no  need to compute in world frame, `sensor_frame`=measurement frame is completely fine in my opinion).
I generalized the force field to a vector, to permit having a tilted world in which gravity is not along -z.

I would still like to add a test using a `param.yaml` + maybe some transforms to fully evaluate the filter in real conditions. The current tests were not sufficient to spot the wrench "transform" errors, because they set up identity transforms which are a special case.

src/control_filters/gravity_compensation.cpp
```diff
@@ -0,0 +1,107 @@
...
+      data_out.header.stamp = data_in.header.stamp;  // only copy the timestamp
+    }
+    else
+    {
+      data_out.header = data_in.header;  // keep the same header and same frame_id
+    }
+    transform_data_out_sensor_ = p_tf_Buffer_->lookupTransform(
+        data_out.header.frame_id, parameters_.sensor_frame, rclcpp::Time());
+    // rotation only (because gravity is a field) from world (gravity frame) to sensor frame
+    rot_sensor_world_ = p_tf_Buffer_->lookupTransform(
```
destogl, line 59: Are you sure this is correct?

test/control_filters/test_gravity_compensation.cpp
```diff
@@ -0,0 +1,100 @@
...
+TEST_F(GravityCompensationTest, TestGravityCompensationParameters)
+{
+  std::shared_ptr<filters::FilterBase<geometry_msgs::msg::WrenchStamped>> filter_ =
+    std::make_shared<control_filters::GravityCompensation<geometry_msgs::msg::WrenchStamped>>();
+
+  double gravity_acc = 9.81;
+  double mass = 5.0;
+  node_->declare_parameter("world_frame", "world");
+  node_->declare_parameter("sensor_frame", "sensor");
+  node_->declare_parameter("CoG.force", std::vector<double>({0.0, 0.0, -gravity_acc * mass}));
```
destogl, line 42: We shouldn't do this. We have generate_parameter_library to take care of this things. So no declaration of setting of parameters, but using generate_parameter_library that sets the params when node is started.

destogl: For differnet cases simply use different `*.yaml` and test files

test/control_filters/test_gravity_compensation.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "rclcpp/rclcpp.hpp"
+
+namespace
+{
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("test_gravity_compensation");
+}  // namespace
+
+// TODO(destogl): do this
+// subclassing and friending so we can access member variables
+
```
destogl, line 33:
```suggestion
```

destogl (changes requested):
The calculations look too complex to me – at least to the older version as I can remember. Is everything necessary?

christophfroehlich (commented):
Could we add a short documentation here please? Some (pseudo-) formulas to describe the mathematics would be great, otherwise it is very hard to review.
Maybe in a separate readme.[md,rst] together with some high-level description of what the filter is good for.

test/control_filters/test_gravity_compensation.cpp
guihomework: hum, ok. I did not know this change to `generate_parameter_library` had to also be applied to the tests which are quite flexible with node parameter declaration. 
Indeed, in other PRs where I had to create tests with a YAML, the values in the YAML have to match the values in the test (to verify the values were loaded correctly by code). This means 2 files have to be edited when setting the parameters, to ensure everything matches. It is easier when params and verified in one place. 
 Maybe I could move to "set_parameter" only instead of declaring. But I think this might be complex or impossible because param declaration happens in the filter_chain, with the node passed to it, and should already have the params overriden there. To override the params in the node, they must be declared isn't it ? this is a chicken-egg problem.

src/control_filters/gravity_compensation.cpp
guihomework: for the rotation, well, if I use the full transform, the -mg force along z world gets transformed from the world to sensor frame, generating a force matching -mg in the sensor frame, but also the cross product of the relative distance between sensor and world with the -mg, creating a torque. Gravity is not creating such a torque at the CoG. The only cross product is the distance of CoG to center of measurement with the simple gravity force at CoG (but oriented correctly).
 The same calculation is done there https://github.com/ros-controls/ros2_controllers/blob/master/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp#L318

guihomework:
> Could we add a short documentation here please? Some (pseudo-) formulas to describe the mathematics would be great, otherwise it is very hard to review. Maybe in a separate readme.[md,rst] together with some high-level description of what the filter is good for.

Sure, this could permit to avoid others not understanding immediately what was going on either, redoing all the maths.

guihomework:
> The calculations look too complex to me – at least to the older version as I can remember. Is everything necessary?
> 
> Also, tests should not declare parameters if possible, but generate_parameter_library should take care of it

I think I will clarify with a tiny drawing in a readme. I would say the code I started from was a special case, and did not work in all cases, especially if the force torque measurement does not happen in a frame that the kinematics knows about (not talking about lookup here).
One example was in case the force torque sensor is mounted at the robot flange, but the measurement happens 5 cm further (thickness of the sensor). So one needs to consider this transform, which adds one transform. But I simplified (removed one transform from Cog to world) the original code to not have to move the computation in world frame and back. World frame is just where gravity is defined. Computation do not have to be there I believe.

include/control_filters/gravity_compensation.hpp
```diff
@@ -0,0 +1,131 @@
...
+template <typename T>
+GravityCompensation<T>::~GravityCompensation()
+{
+}
+
+template <typename T>
+bool GravityCompensation<T>::configure()
+{
+  clock_ = std::make_shared<rclcpp::Clock>(RCL_SYSTEM_TIME);
+  p_tf_Buffer_.reset(new tf2_ros::Buffer(clock_));
```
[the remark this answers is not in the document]
guihomework: Unless the filter_base is changed, I don't think the filter has any access to the node (only its logger and parameter interfaces). Maybe the filter should get a time and dt passed when updated (as controllers do)

guihomework:
> Could we add a short documentation here please? Some (pseudo-) formulas to describe the mathematics would be great, otherwise it is very hard to review. Maybe in a separate readme.[md,rst] together with some high-level description of what the filter is good for.

I added a README.md inside the source folder, as the general README does not even describe the PID etc...

> For differenet cases simply use different `*.yaml` and test files

I will try address this in some near future (also for #152)

christophfroehlich:
> I added a README.md inside the source folder, as the general README does not even describe the PID etc...

Great thx, I already realized the lack of any docs here. -> #154  ;)

I'll have a look the next days.

test/control_filters/test_gravity_compensation.cpp
guihomework: I did the change. I am not sure this is was really needed, but at least there is no "visible" declaration.

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
+# Control Filters
+
+## Available filters
+
+* Gravity Compensation: implements a gravity compensation algorithm, removing the gravity component from the incoming data (Wrench).
+*
```
christophfroehlich, line 6:
```suggestion

```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
+# Control Filters
```
christophfroehlich, line 1:
```suggestion
# Control filters
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Available filters
+
+* Gravity Compensation: implements a gravity compensation algorithm, removing the gravity component from the incoming data (Wrench).
+*
+
+
+## Gravity Compensation filter
```
christophfroehlich, line 9:
```suggestion
## Gravity compensation filter
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+ This filter implements a gravity compensation algorithm, applied to an `data_in wrench`, computed at a `sensor frame` in which the center of gravity (CoG) of the to-be-compensated mass is known.
+
+ The filter relies on ROS TF, and might fail if transforms are missing.
```
christophfroehlich, line 13:
```suggestion
 The filter relies on tf2, and might fail if transforms are missing.
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Available filters
+
+* Gravity Compensation: implements a gravity compensation algorithm, removing the gravity component from the incoming data (Wrench).
+*
+
+
+## Gravity Compensation filter
+
+ This filter implements a gravity compensation algorithm, applied to an `data_in wrench`, computed at a `sensor frame` in which the center of gravity (CoG) of the to-be-compensated mass is known.
```
christophfroehlich, line 11:
```suggestion
 This filter implements a gravity compensation algorithm, applied to an `data_in` wrench, computed at a `sensor_frame` in which the center of gravity (CoG) of the to-be-compensated mass is known.
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+ The filter relies on ROS TF, and might fail if transforms are missing.
+
+ Note that, for convenience, the filter can perform additional frame changes if data_out frame id is given.
+
+### required parameters
```
christophfroehlich, line 17:
```suggestion
### Required parameters
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+* `world_frame` (&Rscr;<sub>w</sub>): frame in which the `CoG.force` is represented.
+* `sensor_frame` (&Rscr;<sub>s</sub>): frame in which the `CoG.pos` is defined
+* `CoG.pos` (p<sub>s</sub>): position of the CoG of the mass the filter should compensate for
+* `CoG.force` (g<sub>w</sub>): constant (but updatable) force of gravity at the Cog (typically m.G), defined along axes of the `world_frame`
+
+### algorithm
```
christophfroehlich, line 24:
```suggestion
### Algorithm
```

src/control_filters/gravity_compensation_filter_parameters.yaml
```diff
@@ -0,0 +1,34 @@
+gravity_compensation_filter:
+  world_frame: {
+    type: string,
+    # default_value: world,
```
christophfroehlich, line 4:
```suggestion
```

src/control_filters/gravity_compensation_filter_parameters.yaml
```diff
@@ -0,0 +1,34 @@
...
+    description: "fixed world frame in which the constant field-induced force is given",
+    read_only: true,
+    validation: {
+      not_empty<>: []
+    },
+  }
+  sensor_frame: {
+    type: string,
+    # default_value: ft_sensor,
```
christophfroehlich, line 13:
```suggestion
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Given
+* above-required parameters,  &Rscr;<sub>w</sub>, &Rscr;<sub>s</sub>, p<sub>s</sub>, g<sub>w</sub>
+* `data_in`, a wrench &Fscr;<sub>i</sub> = {f<sub>i</sub>, &tau;<sub>i</sub>} represented in `data_in.frame` &Rscr;<sub>i</sub>
+* access to TF homogeneous transforms:
+  * T<sub>si</sub> from &Rscr;<sub>i</sub> to &Rscr;<sub>s</sub>
+  * T<sub>sw</sub> from &Rscr;<sub>w</sub> to &Rscr;<sub>s</sub>
+  * T<sub>os</sub> from &Rscr;<sub>s</sub> to &Rscr;<sub>o</sub>
+
+Compute `data_out` compensated wrench &Fscr;c<sub>o</sub> = {fc<sub>o</sub>, &tau;c<sub>o</sub>} represented in `data_out.frame` &Rscr;<sub>o</sub> if given, or `data_in.frame` &Rscr;<sub>i</sub> otherwise, with equations:
```
christophfroehlich, line 34:
```suggestion
Compute `data_out` compensated wrench &Fscr;c<sub>o</sub> = {fc<sub>o</sub>, &tau;c<sub>o</sub>} represented in the `data_out` frame &Rscr;<sub>o</sub> if given, or `data_in` frame &Rscr;<sub>i</sub> otherwise, with equations:
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+* `world_frame` (&Rscr;<sub>w</sub>): frame in which the `CoG.force` is represented.
+* `sensor_frame` (&Rscr;<sub>s</sub>): frame in which the `CoG.pos` is defined
+* `CoG.pos` (p<sub>s</sub>): position of the CoG of the mass the filter should compensate for
+* `CoG.force` (g<sub>w</sub>): constant (but updatable) force of gravity at the Cog (typically m.G), defined along axes of the `world_frame`
+
+### algorithm
+
+Given
+* above-required parameters,  &Rscr;<sub>w</sub>, &Rscr;<sub>s</sub>, p<sub>s</sub>, g<sub>w</sub>
+* `data_in`, a wrench &Fscr;<sub>i</sub> = {f<sub>i</sub>, &tau;<sub>i</sub>} represented in `data_in.frame` &Rscr;<sub>i</sub>
```
christophfroehlich, line 28:
```suggestion
* `data_in`, a wrench &Fscr;<sub>i</sub> = {f<sub>i</sub>, &tau;<sub>i</sub>} represented in the `data_in` frame &Rscr;<sub>i</sub>
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+* access to TF homogeneous transforms:
```
christophfroehlich, line 29:
```suggestion
* access to tf2 homogeneous transforms:
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+Compute `data_out` compensated wrench &Fscr;c<sub>o</sub> = {fc<sub>o</sub>, &tau;c<sub>o</sub>} represented in `data_out.frame` &Rscr;<sub>o</sub> if given, or `data_in.frame` &Rscr;<sub>i</sub> otherwise, with equations:
+
+&Fscr;c<sub>o</sub> = T<sub>os</sub>.&Fscr;c<sub>s</sub>,
+
+
+with &Fscr;c<sub>s</sub> = {fc<sub>s</sub>, &tau;c<sub>s</sub>}  the compensated wrench in `sensor_frame` (common frame for computation)
+
+and,
+
+fc<sub>s</sub> = f<sub>s</sub> - T<sub>sw</sub>.g<sub>w</sub>
```
christophfroehlich, line 43:
```suggestion
fc<sub>s</sub> = f<sub>s</sub> - T<sub>sw</sub>g<sub>w</sub>
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+its force  and,
+
+&tau;c<sub>s</sub> = &tau;<sub>s</sub> - p<sub>s</sub> x (T<sub>sw</sub>.g<sub>w</sub>)
```
christophfroehlich, line 47:
```suggestion
&tau;c<sub>s</sub> = &tau;<sub>s</sub> - p<sub>s</sub> x (T<sub>sw</sub>g<sub>w</sub>)
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+its torque, and
+
+&Fscr;<sub>s</sub>  = T<sub>si</sub>.&Fscr;<sub>i</sub>
```
christophfroehlich, line 51:
```suggestion
&Fscr;<sub>s</sub>  = T<sub>si</sub>&Fscr;<sub>i</sub> = {f<sub>s</sub>, &tau;<sub>s</sub>}
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+the full transform of the input wrench &Fscr;<sub>i</sub> to sensor frame &Rscr;<sub>s</sub>
+
+Remarks :
+* a full vector is used for gravity force, to not impose gravity to be only along z of `world_frame`.
+* `data_in.frame` is usually equal to `sensor_frame`, but could be different since measurement of wrech might occur in another frame. Ex: measurements are at the **FT sensor flange** = `data_in.frame`, but CoG is given in **FT sensor base** = `sensor_frame` (=frame to which it is mounted on the robot), introducing an offset (thichkess of the sensor) to be accounted for.
```
christophfroehlich, line 57:
```suggestion
* `data_in` frame is usually equal to `sensor_frame`, but could be different since measurement of wrench might occur in another frame. E.g.,  measurements are at the **FT sensor flange** = `data_in` frame, but CoG is given in **FT sensor base** = `sensor_frame` (=frame to which it is mounted on the robot), introducing an offset (thickness of the sensor) to be accounted for.
```

src/control_filters/README.md
```diff
@@ -0,0 +1,59 @@
...
+* `data_out.frame` is usually `data_in.frame`, but for convenience, can be set to any other useful frame. Ex: Wrench expressed in a `control_frame` for instance center of a gripper.
```
christophfroehlich, line 58:
```suggestion
* `data_out` frame is usually `data_in` frame, but for convenience, can be set to any other useful frame. E.g., wrench expressed in a `control_frame` like the center of a gripper.
```

christophfroehlich (approved):
LGTM

Thanks for the docs, sounds reasonable and according to the actual implementation.

guihomework:
Thanks a lot @christophfroehlich for the review and sorry for fixing spell-checking, I had forgotten to install a spell checker in VS Code, my former editor had one integrated.

Now, only the issue about removing clock access is open I believe and is not straight forward to solve. @destogl can we live with this ?

christophfroehlich (approved):

include/control_filters/gravity_compensation.hpp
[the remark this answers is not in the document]
bmagyar: maybe for a follow-up PR, you could augment the filterbase to also store a clock interface

test/control_filters/test_gravity_compensation.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  ASSERT_EQ(out.wrench.torque.x, 10.0);
+  ASSERT_EQ(out.wrench.torque.y, 0.0);
+  ASSERT_EQ(out.wrench.torque.z, 0.0);
+
+  out.header.frame_id = "base";
+  // should fail due to missing transform for desired output frame
+  ASSERT_FALSE(filter_->update(in, out));
+
+  // TODO(guihomework) Add a test with real lookups
```
bmagyar, line 103: please create a follow-up issue, describing precisely what you mean by this comment and link it here instead of the text TODO

bmagyar (approved):

test/control_filters/test_gravity_compensation.hpp
guihomework: I will also create a follow-up for an improved member variable

guihomework:
I just added one export that solves an issue when building in `symlink-install` 
```CMake Error at /opt/ros/rolling/share/ament_cmake_core/cmake/symlink_install/ament_cmake_symlink_install_targets.cmake:81 (message):
  ament_cmake_symlink_install_targets()
  'gravity_compensation_filter_parameters' is an interface library - there's
  nothing to symlink install.  Perhaps you forgot to add EXPORT or INCLUDES
  DESTINATION?
```

include/control_filters/gravity_compensation.hpp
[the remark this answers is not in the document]
christophfroehlich: https://github.com/ros-controls/control_toolbox/issues/162

christophfroehlich (approved):

control_toolbox/src/control_filters/gravity_compensation.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+template <>
+bool GravityCompensation<geometry_msgs::msg::WrenchStamped>::update(
+  const geometry_msgs::msg::WrenchStamped & data_in, geometry_msgs::msg::WrenchStamped & data_out)
+{
+  if (!this->configured_)
+  {
+    RCLCPP_ERROR_SKIPFIRST_THROTTLE((*logger_), *clock_, 2000, "Filter is not configured");
+    return false;
+  }
```
saikishor, line 32:
```suggestion
```

saikishor (commented):
Looks good. I haven't gone through the math

control_toolbox/src/control_filters/README.md
```diff
@@ -4,9 +4,65 @@ Implement filter plugins for control purposes as https://index.ros.org/r/filters
...
 ## Available filters
 
+* Gravity Compensation: implements a gravity compensation algorithm, removing the gravity component from the incoming data (Wrench).
 * Low Pass: implements a low-pass filter based on a time-invariant [Infinite Impulse Response (IIR) filter](https://en.wikipedia.org/wiki/Infinite_impulse_response), for different data types (doubles or wrench).
 * Exponential Filter: Exponential filter for double data type.
 
+
+## Gravity compensation filter
+
+ This filter implements a gravity compensation algorithm, applied to an `data_in` wrench, computed at a `sensor_frame` in which the center of gravity (CoG) of the to-be-compensated mass is known.
```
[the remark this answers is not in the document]
guihomework: Hi, as the original author of the PR, I still get the notifications, and although I am not into ROS 2 anymore, and don't know if changes to clock constructor (ROS or SYSTEM times) and others are required, I still can report on english text change.
if we simplify the sentence we see that the suggested fix does now work grammatically.
see "This filter implements an algorithm, applied to a wrench, computed at a frame in which the center of gravity (CoG) known mass." 
I can understand that sentence is not simple, so what about rephrasing :

This filter implements an algorithm compensating for the gravity forces acting at the center of gravity (CoG) of a known mass, computed at a `sensor_frame` and applied to a `data_in` wrench.

control_toolbox/src/control_filters/gravity_compensation_filter_parameters.yaml
```diff
@@ -0,0 +1,32 @@
...
+        fixed_size<>: 3
+      },
+    }
+    force: {
+        type: double_array,
+        description: "Specifies the constant field-induced force applied at the CoG of tool defined in fixed world frame, e.g [0, 0, -mass * 9.81].",
+        validation: {
+          fixed_size<>: 3
+        },
+      }
```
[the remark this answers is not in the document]
guihomework: :+1:  Sure, this leads to some more lines in the code to store the scalar product of the field vector and the mass. The only advantage of a full "force-field vector" as a parameter, is that it could be any force-field (not only gravity but other, even if this is rare to have other fields that this one, and the filter is called gravity compensation...)

control_toolbox/src/control_filters/README.md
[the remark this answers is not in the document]
saikishor: Sounds fine to me :)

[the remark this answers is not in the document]
saikishor: ROS time would make more sense because when the use_sim_time parameter is not set, it is nothing but SYSTEM time and when the parameter is set, it uses the simulation clock

control_toolbox/src/control_filters/gravity_compensation_filter_parameters.yaml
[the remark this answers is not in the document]
saikishor: I understand, but having it separated has its own advantages. For instance, depending on the inclination of the robot, if I have to change the gravity vector, i can simply change it without knowing anything of the mass of the gripper etc. With the current method, I'll have to know about the gravity vector alignment as well as the mass of the gripper, which is the information I'll have to duplicate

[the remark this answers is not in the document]
guihomework: I unfortunately don't have a setup to easily update and test code, but sure your explanation makes sense as well.

control_toolbox/src/control_filters/README.md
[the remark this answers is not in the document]
guihomework: Here the clock is only used to throttle the messages for a human to read a warnings/errors in a slow fashion (instead of maybe 1 kHz of a controller update, so following the sim time makes no sense. Suppose you have a 10 times faster than realtime sim, this means the console would be flooded 10 times more than expected, by this 5 second throttled messages. I think SYSTEM time is what I aimed for due to logs making sense at human rate and human rate is SYSTEM time. I would agree with ROS time if there were any computation related to the clock, which is not the case as far as I can see where the clock_ is used.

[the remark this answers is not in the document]
guihomework: You are right, I did a search for clock_ but missed the one passed to tfBuffer, it was right below your change, really no idea why I missed it. 2 years and I forgot things.  Yes tf::transforms require the same clock as the transforms so ROS time is even mandatory. Probably due to a powerful computer I had at the time, that was running gazebo in realtime, I did not notice a the problem.
So good catch, thanks.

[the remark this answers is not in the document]
christophfroehlich:
```suggestion
This filter implements an algorithm compensating for the gravity forces acting at the center of gravity (CoG) of a known mass, computed at a `sensor_frame` and applied to a `data_in` wrench.
```

control_toolbox/src/control_filters/gravity_compensation.cpp
christophfroehlich: we have this check in the other filters as well:
https://github.com/ros-controls/control_toolbox/blob/3ed348a6736d763b0170d0bb3ee3d8fe0b0e42a3/control_toolbox/include/control_filters/exponential_filter.hpp#L110-L113

But in the other filters we throw an exception, instead of logging. Should we change it?

saikishor: Sure. Let's do as other filters :)

christophfroehlich (dismissed):

control_toolbox/src/control_filters/gravity_compensation.cpp
christophfroehlich: done

guihomework: What I can say is that when I was asked to integrate the parameters with the generate_parameter_library, I did try to change the minimum possible while checking the code was doing what it was supposed to (hence fixing also the computation at some point), but the original PR had a return with false so I kept it.  
Having an "OROCOS" background, where there is a state machine, a component could not run if it was not configured properly, so the run/update phase was not supposed to check again if the controller was configured, but the whole idea was to not crash/end the whole control loop due to deployment (load unload) at runtime.  I forgot if the controller manager correctly catches those throw.  If there is a mechanism capturing, then a thrown might be fine.

christophfroehlich: The controller_manager handles this (we removed those checks in the controllers), but this filter could be used stand-alone as well. Then it's the responsibility of the user to take care of that

saikishor: We decided to throw an exception because if someone forgot to configure the plugin and tried to run update directly, it shouldn't continue returning false as it is not functional or recoverable in this state, so we thought throwing an exception is the right idea

guihomework: thanks for the clarification. I am glad to see this work is going to be integrated. Always cool to see something being used.

control_toolbox/src/control_filters/gravity_compensation_filter_parameters.yaml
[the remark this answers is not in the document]
christophfroehlich: I tend towards the argument of Sai, and fixed that quickly. I changed the `CoG` parameter name to `tool`, because it was not semantically correct any more. Other suggestions?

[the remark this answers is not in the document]
saikishor: @christophfroehlich can I do this change?

Do you agree with this?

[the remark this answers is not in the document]
christophfroehlich: too late ;)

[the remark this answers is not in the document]
saikishor: LOL! You are too fast 😂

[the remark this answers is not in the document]
guihomework: May I mention that, for me, user of robots with tools that changed often, I had 2 parameters to set on some kuka robot with gravity compensation integrated. One was the center of mass position (of the tool or what ever else), and the second was one was the tool frame offset/rotation.  For one of our gripper, the tool frame was somewhere else than the CoG. I would say that only "tool" pos is confusion for me. Why is CoG not semantically correct ?  tool/com_pos would be appropriate.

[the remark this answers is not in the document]
christophfroehlich: because `CoG.mass` or `CoG.gravity_field` sounds odd to me. Imho, "center of gravity" can't have attributes other than position/orientation.

saikishor (approved):
The changes look good to me. 

I'm just not sure about using lookup transform directly in the RT loop. We should take a look at it and improve it.

control_toolbox/src/control_filters/gravity_compensation_filter_parameters.yaml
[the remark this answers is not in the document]
guihomework: Agreed, change from 'CoG' to 'tool' is better, but 'tool.pos' might be misleading and might requires clarification. Maybe I am biased by parameters on ROS being all on the parameters server and loaded globally where all parameters could be mixed up in the same namespace, here the parameters are per node (even more fine grained as it is for the filters here) in ROS 2 so probably less likely to be misleading ?
I just wanted to raise this point from the user perspective.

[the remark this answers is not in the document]
christophfroehlich: ah, you are right. I changed it to `tool.CoG`. Is this clearer now?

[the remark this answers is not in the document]
guihomework: perfect. thanks.

saikishor (approved):

guihomework:
> I'm just not sure about using lookup transform directly in the RT loop. We should take a look at it and improve it.

There are thoughts and some explanation on the reason of using lookup transform in some out-dated/hidden remarks in the PR, but the main issue was that the computation requires to know something the kinematics/controllers did not know in the use case we had, especially if the real life sensor had a thickness, and the 'sensor_frame' was not at the flange (end-effector frame). 
The second issue was probably that the filter just received a wrench, in some frame, and had to deal with it and convert the gravity field in that frame correctly to compensate for it.  

I think it was decided to let it as is, but I agree it is not nice to have lookup transforms in an RT loop. I think it requires passing more information to the update loop if the lookup must be removed, so that the controller itself, having access to the kinematics can compute it, but it might also require the sensor position and CoG position of the mass to pre-compute those transforms for the filter. Not so trivial.

saikishor:
@guihomework yes, I understood that it was used due to the end effector transformation. Maybe, it would be nice to have an update method with this extra transformation arg, so if a controller can compute it, then it can use this method to avoid it. This can be a future work

christophfroehlich (approved):
Thanks for the initial work and the discussion, I think we can merge this now!

PR #180, opened 2024-02-11
[PID] Add support for saving i-term when PID is reset

destogl:
Saves integral term on reset to maintain steady state offset output.
This is required for controlling of system without integral part to achieve correct steady-state output.
If intergral term is reset for some time there will be incorrect output in steady-state. This features avoids such situations and makes btter performance.


olivier-stasse (approved):
LGTM

christophfroehlich (changes requested):
There is no test if this parameter has the desired effect. Could you please add any? e.g., to `gainSettingCopyPIDTest` or add a new one..

christophfroehlich (approved):
I added a test for this functionality to `integralOnlyTest`

christophfroehlich (approved):

saikishor (changes requested):
The changes in the PR look fine to me, just some minor changes here

Apart from that, just a conceptual question. I'm not sure if the `save_i_term` should be the part of the Gains itself or it should be a parameter to the `reset` method alone. If everyone is happy with the original choice, I'm fine with it.

christophfroehlich (approved):
I applied the changes requested by @saikishor and had a focus on not breaking ABI -> check is green -> we can merge that into jazzy imho.

include/control_toolbox/pid.hpp
```diff
@@ -201,6 +211,18 @@ class Pid
    */
```
saikishor, line 211:
```suggestion
   * @note The integral term is not retained and it is reset to zero
   */
```

Just to document

include/control_toolbox/pid_ros.hpp
```diff
@@ -113,6 +128,13 @@ class PidROS
    */
```
saikishor, line 128:
```suggestion
   * @note The integral term is not retained and it is reset to zero
   */
```

include/control_toolbox/pid_ros.hpp
```diff
@@ -113,6 +128,13 @@ class PidROS
...
   void reset();
 
+  /*!
+   * \brief Reset the state of this PID controller
+   *
+   * \param save_iterm boolean indicating if integral term is retained on reset()
+   */
+  void reset(bool save_iterm = false);
```
saikishor, line 136:
```suggestion
  void reset(bool save_iterm);
```

It is better to remove the default arg, I'm not sure how the compiler can differentiate between the non parsed one and the default param one

src/pid.cpp
```diff
@@ -76,12 +82,30 @@ void Pid::initPid(double p, double i, double d, double i_max, double i_min, bool
...
   p_error_ = 0.0;
-  i_error_ = 0.0;
   d_error_ = 0.0;
   cmd_ = 0.0;
+
+  // If last integral error is already zero, just return
+  if (std::abs(i_error_) < std::numeric_limits<double>::epsilon())
+  {
+    return;
+  }
```
saikishor, line 100: Do we need this check?

saikishor (approved):
In general, LGTM

include/control_toolbox/pid_ros.hpp
christophfroehlich: it did not complain although I have used both versions in the tests. but it is not really necessary here.

src/pid.cpp
christophfroehlich: maybe not? it does not take more effort to just zero it.

christophfroehlich, line 100:
```suggestion
```

christophfroehlich: in the earlier version we had to use the RT buffer here, that's why

saikishor: Got it

saikishor (approved):

christophfroehlich (approved):

PR #192, opened 2024-03-02
Add standalone version of LPF (#164)

### LINKED ISSUE #164: LowPassFilter accepts parameters only from node
I wanted to use `LowPassFilter`, but it seems to me that there is no way to pass parameters via API (like `PID` constructor or `setGains()` call).

Wouldn't be better to make `LowPassFilter` support parameter passing via API and a corresponding `LowPassFilterROS` querying parameters from configuration instead? 
Or just modifying LowPassFilter to support both scenarios... 

I would contribute a PR following your feedbacks of course.

roncapat:
Closes #164.
Test passes locally.

The old `LowPassFilter` class is now `LowPassFilterRos`.
API-breaking since with this patch `LowPassFilter` is a parent class without ROS parameters.

@christophfroehlich at least for now I wanted to reflect the same difference of `Pid` and `PidROS`. But we need to devise a different naming strategy, or this patch will break a lot of perfectly fine code (will require people to swap `LowPassFilter` with `LowPassFilterRos` in their projects).


christophfroehlich:
Sorry for the late reply. I see two options:

- Use a consistent nomenclature and break API with Jazzy (we would have to branch this repo then)
- Deprecate the old `LowPassFilter` with the hint to replace it with `LowPassFilterRos` and call the base filter `LowPassFilterBase` or similar. This could be then rewritten in a future release to converge back to the consistent nomenclature.

What do you think?

roncapat:
> Sorry for the late reply. I see two options:
> 
> - Use a consistent nomenclature and break API with Jazzy (we would have to branch this repo then)
> - Deprecate the old `LowPassFilter` with the hint to replace it with `LowPassFilterRos` and call the base filter `LowPassFilterBase` or similar. This could be then rewritten in a future release to converge back to the consistent nomenclature.
> 
> What do you think?

Yeah I believe we should slowly make nomenclature consistent, I like better solution 2.

BTW, why are there no 'humble', 'iron', 'rolling' branches and so on in this repo?

roncapat:
I will update the PR as soon as possible implementing what you proposed as option 2.

christophfroehlich:
> BTW, why are there no 'humble', 'iron', 'rolling' branches and so on in this repo?

because the ros2-master branch is released to all distros, there was no need for branching yet

christophfroehlich:
@roncapat do you still plan to work on this PR?

roncapat:
@christophfroehlich sorry but at least for now I'm not able to finish this one due of lack of time. Of course more than happy if someone else can pick this up.

PR #200, opened 2024-04-24
[CI] Specify runner/container images and add Jazzy jobs

christophfroehlich:
- Necessary changes to make https://github.com/ros-controls/ros2_control_ci/pull/53 work for noble.
- Add jobs for Jazzy

Jobs for jazzy-main are failing, but jazzy-testing works fine already.


.github/workflows/ci-pre-commit.yml
```diff
@@ -10,6 +10,13 @@ jobs:
...
+        ROS_DISTRO: [iron, humble]
     with:
       ros_distro: ${{ matrix.ROS_DISTRO }}
+      os_name: ubuntu-22.04
+
+  pre-commit-rolling:
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-pre-commit.yml@master
+    with:
+      ros_distro: rolling
+      container: ubuntu:24.04
```
fmauch, line 22: I think that would be a nice change. Then, we only need to add another matrix configuration for Jazzy

```suggestion
        include:
          - ROS_DISTRO: humble
            OS_NAME: 22.04
          - ROS_DISTRO: iron
            OS_NAME: 22.04
          - ROS_DISTRO: rolling
            OS_NAME: 24.04
    with:
      ros_distro: ${{ matrix.ROS_DISTRO }}
      os_name: ${{ matrix.OS_NAME }}
```

christophfroehlich: we need a different input os_name vs container. but this could work anyways.

fmauch: Oh, I missed that. Can we access matrix variables that aren't set?

christophfroehlich: do you have a feeling if it is a bad idea to just always specify a container instead of running the job "locally" on the runner? then I could just specify the container and let os_name be ubuntu-latest

fmauch: I think that should be fine. I would expect runners to cache frequently used images.

fmauch (approved):
That's nice, thank you for trying that out!

fmauch (approved):
Seems good to me.

PR #212, opened 2024-10-13
Move speed limiter from ros2_control repo

christophfroehlich:
I thought it would be beneficial to reuse the SpeedLimiter of the diff_drive controller in the steering_controllers_library.

The SpeedLimiter might be useful for other controllers as well, so I suggest to add it to the control_toolbox repo.
But how can we deal with the change of the license from Apache 2 to BSD-3-Clause in this repo? Other suggestions?

Additionally, I added some tests because there were none.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,9 @@ if(BUILD_TESTING)
   ament_add_gmock(pid_tests test/pid_tests.cpp)
   target_link_libraries(pid_tests control_toolbox)
 
+  ament_add_gmock(speed_limiter_tests test/speed_limiter.cpp)
+  target_link_libraries(speed_limiter_tests control_toolbox)
+
   ament_add_gtest(pid_parameters_tests test/pid_parameters_tests.cpp)
   target_link_libraries(pid_parameters_tests control_toolbox)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ add_library(control_toolbox SHARED
   src/pid.cpp
   src/sine_sweep.cpp
   src/sinusoid.cpp
+  src/speed_limiter.cpp
 )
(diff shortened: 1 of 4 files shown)


bmagyar:
Very simple. We should relicense all bsd to apache.

test/control_filters/test_low_pass_filter.cpp
```diff
@@ -65,6 +65,7 @@ TEST_F(LowPassFilterTest, TestLowPassFilterThrowsUnconfigured)
     ASSERT_THROW(filter_->update(in, out), std::runtime_error);
 }
 
+// TODO(christophfroehlich) why does this test fail sometime?
```
destogl, line 68:
```suggestion
```

bmagyar (approved):

include/control_toolbox/rate_limiter.hpp
```diff
@@ -0,0 +1,306 @@
...
+   * \note
+   * If max_* values are NAN, the respective limit is deactivated
+   * If min_* values are NAN, defaults to -max if unspecified
+   * If min_first_derivative_pos/max_first_derivative_neg values are NAN, symmetric limits are used
+   */
+  void set_params(
+    T min_value = NAN, T max_value = NAN,
+    T min_first_derivative_neg = NAN, T max_first_derivative_pos = NAN,
+    T min_first_derivative_pos = NAN, T max_first_derivative_neg = NAN,
+    T min_second_derivative = NAN, T max_second_derivative = NAN);
```
destogl, line 115: We should use the C++ `NaN` version and not C version.

Basically, the `NaN` should be changed to `std::numeric_limits<double>::quiet_NaN()` - and probably add `#include <limits>` file.

christophfroehlich: done

PR #219, opened 2024-11-02
Add few compiler warning flags to error 

christophfroehlich:
Similar to https://github.com/ros-controls/ros2_controllers/pull/961


src/pid.cpp
```diff
@@ -155,7 +155,7 @@ double Pid::computeCommand(double error, double error_dot, uint64_t dt)
   p_term = gains.p_gain_ * p_error_;
 
   // Calculate the integral of the position error
-  i_error_ += (dt / 1e9) * p_error_;
+  i_error_ += (static_cast<double>(dt) / 1e9) * p_error_;
```
saikishor, line 158: @christophfroehlich the changes of this PR itself looks good

I'm wondering if we should add another method with dt as double?

christophfroehlich: this changes ABI, right? should we really do this back to humble?

saikishor: Adding a new method should be any API/ABI as far as I know.
May be it's worth right?. So, we don't have to always pass the dt in nanoseconds

christophfroehlich: let's do that in a different PR

saikishor: Sure. Just bringing the point

saikishor (approved):
LGTM

PR #220, opened 2024-11-04
Change license to Apache-2

christophfroehlich:
We already have a mix of BSD-3-Clause and Apache-2 licenses in the file. I suggest changing the main LICENSE to Apache, but keep the old one, too.
There are several references online doing so, and github theoretically supports several licenses per repo: https://github.blog/open-source/open-source-license-usage-on-github-com/

btw: I also updated the maintainers in the package.xml

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,25 +1,202 @@
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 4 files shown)


bmagyar (approved):

saikishor (approved):
Thanks for the change!

PR #222, opened 2024-11-04
Add standalone version of LPF

### LINKED ISSUE #164: LowPassFilter accepts parameters only from node

christophfroehlich:
Replaces #192 and closes #164

Changes on top:

- I changed the base class to be independent of `filters`
- Used the old class/file name and moved the new one to `control_toolbox` namespace.

The filter plugins should still work as before without any changes (see the test files).


christophfroehlich:
@roncapat what do you think?

include/control_filters/low_pass_filter.hpp
```diff
@@ -196,55 +154,36 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
...
-
-  data_out.wrench.force.x = msg_filtered[0];
-  data_out.wrench.force.y = msg_filtered[1];
-  data_out.wrench.force.z = msg_filtered[2];
-  data_out.wrench.torque.x = msg_filtered[3];
-  data_out.wrench.torque.y = msg_filtered[4];
-  data_out.wrench.torque.z = msg_filtered[5];
-
-  // copy the header
-  data_out.header = data_in.header;
```
saikishor, line 222: May be the update method specific to geometry_msgs Wrench should be removed?. It kinda has same content of templated one right?

saikishor (changes requested):

include/control_filters/low_pass_filter.hpp
christophfroehlich: the math is the same, but here it uses the Eigen matrices to do the linear combination. this can't work with the templated method alone?

roncapat:
> @roncapat what do you think?

Nice for me! Thank you 👍🏻

saikishor (approved):

PR #229, opened 2024-11-06
LPF: Throw if calling `udpate` unconfigured

christophfroehlich:
@saikishor raised a good point here: https://github.com/ros-controls/control_toolbox/pull/222#discussion_r1829863265

And I removed the shadowed class variables (#219 fails), I don't see the point in saving them anyways.


saikishor (approved):
Thank you!

PR #231, opened 2024-11-08
Add filter plugin for exponential filter

christophfroehlich:
Similar to `control_filters/LowPassFilter`, the filter itself was already there.

### DIFF
--- a/include/control_filters/exponential_filter.hpp
+++ b/include/control_filters/exponential_filter.hpp
@@ -1,0 +1,133 @@
+// Copyright (c) 2024, AIT Austrian Institute of Technology GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CONTROL_FILTERS__EXPONENTIAL_FILTER_HPP_
+#define CONTROL_FILTERS__EXPONENTIAL_FILTER_HPP_
+
+#include <memory>
+#include <limits>
+#include <string>
+
+#include "filters/filter_base.hpp"
+
+#include "control_toolbox/filters.hpp"
+#include "exponential_filter_parameters.hpp"
+
+namespace control_filters
+{
+
+/***************************************************/
+/*! \class ExponentialFilter
+  \brief A exponential filter class.
+
+  This class implements a low-pass filter for
+  various data types.
+
+  \section Usage
+
+  The ExponentialFilter class is meant to be instantiated as a filter in
+  a controller but can also be used elsewhere.
+  For manual instantiation, you should first call configure()
+  (in non-realtime) and then call update() at every update step.
+
+*/
+/***************************************************/
+
+template <typename T>
+class ExponentialFilter : public filters::FilterBase<T>
+{
+public:
+  /*!
+   * \brief Configure the ExponentialFilter (access and process params).
+   */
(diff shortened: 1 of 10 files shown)


test/control_filters/test_exponential_filter.cpp
```diff
@@ -0,0 +1,65 @@
...
+      in = 0.0;
+    for (int i = 0; i < 100; ++i){
+      ASSERT_TRUE(filter_->update(in, out));
+      calculated = alpha * in + (1 - alpha) * calculated;
+      ASSERT_EQ(calculated, out);}
+}
+
+int main(int argc, char ** argv)
+{
+  ::testing::InitGoogleTest(&argc, argv);
```
bmagyar, line 60:
```suggestion
  ::testing::InitGoogleMock(&argc, argv);
```

bmagyar (approved):

include/control_filters/exponential_filter.hpp
```diff
@@ -0,0 +1,133 @@
...
+    catch (rclcpp::exceptions::ParameterUninitializedException & ex) {
+      RCLCPP_ERROR((*logger_), "Exponential filter cannot be configured: %s", ex.what());
+      parameter_handler_.reset();
+      return false;
+    }
+    catch (rclcpp::exceptions::InvalidParameterValueException & ex)  {
+      RCLCPP_ERROR((*logger_), "Exponential filter cannot be configured: %s", ex.what());
+      parameter_handler_.reset();
+      return false;
+    }
```
saikishor, line 98:
```suggestion
    catch (const std::exception & ex) {
      RCLCPP_ERROR((*logger_), "Exponential filter cannot be configured: %s (type : %s)", ex.what(), typeid(e).name());
      parameter_handler_.reset();
      return false;
    }
```

include/control_filters/exponential_filter.hpp
```diff
@@ -0,0 +1,133 @@
...
+    throw std::runtime_error("Filter is not configured");
+  }
+
+  // Update internal parameters if required
+  if (parameter_handler_->is_old(parameters_))
+  {
+    parameters_ = parameter_handler_->get_params();
+  }
+
+  if (std::isnan(last_smoothed_value))
```
saikishor, line 121:
```suggestion
  if (!std::isfinite(last_smoothed_value))
```

saikishor (approved):
In general looks great.
Thanks for this new filter

include/control_filters/exponential_filter.hpp
saikishor: Similar to https://github.com/ros-controls/ros2_control/blob/53656e4636ba0472a5ee20d3a29f096ff50f6f5d/controller_manager/src/controller_manager.cpp#L544-L551

test/control_filters/test_exponential_filter.cpp
christophfroehlich: I'll move that in a separate issue, because this is used in the other tests as well.

include/control_filters/exponential_filter.hpp
christophfroehlich: Should I add a `catch (...)` as well?

christophfroehlich: I move that also to a different issue, because this pattern is the same for every filter here.

include/control_filters/exponential_filter.hpp
christophfroehlich: isn' this legitimate here as we initialize this variable with `std::numeric_limits<T>::quiet_NaN()`?

include/control_filters/exponential_filter.hpp
saikishor: Sure. I agree. THis can be a different issue. May be we can add catch(..) as well.

include/control_filters/exponential_filter.hpp
saikishor: Sure. it is fine

test/control_filters/test_exponential_filter.cpp
bmagyar: Thanks!

saikishor (approved):

Issue #234, opened 2024-11-15
TestLowPassFilterThrowsUnconfigured times out from time to time


christophfroehlich:
> ```
> 2024-11-15T06:46:33.4191267Z 4: [ RUN      ] LowPassFilterTest.TestLowPassFilterThrowsUnconfigured
> 2024-11-15T06:46:33.4191993Z 4/5 Test #4: test_low_pass_filter .............***Timeout  60.06 sec
> ```
#233

PR #261, opened 2025-01-04
Fix control_filters tests

christophfroehlich:
The problem of #234 was that the executor in the separate thread sometimes just doesn't stop spinning, and the `join()` method hangs. But with these tests, we don't need a spinning node (we just use the parameter and logging interface).

To simplify things, I moved that to a common test_fixture.


test/control_filters/test_filter_util.hpp
```diff
@@ -12,52 +12,40 @@
...
-    if (executor_thread_.joinable())
-    {
-      executor_thread_.join();
-    }
     node_.reset();
   }
 
 protected:
   rclcpp::Node::SharedPtr node_;
   rclcpp::Executor::SharedPtr executor_;
```
saikishor, line 48: Looks good, but if you don't spin the executor, do you still need one?

christophfroehlich: fair point :D

christophfroehlich: I removed it

saikishor (approved):

Issue #257, opened 2025-01-01
Update pluginlib exception handling


christophfroehlich:
```suggestion
    catch (const std::exception & ex) {
      RCLCPP_ERROR((*logger_), "Exponential filter cannot be configured: %s (type : %s)", ex.what(), typeid(e).name());
      parameter_handler_.reset();
      return false;
    }
```

_Originally posted by @saikishor in https://github.com/ros-controls/control_toolbox/pull/231#discussion_r1900440873_

PR #263, opened 2025-01-07
Update plugin lib exception handling

Juliaj:
Address https://github.com/ros-controls/control_toolbox/issues/257

@christophfroehlich , regarding to catch(...), I left it out since I am not sure what would be useful to log. I'm open to add it if you could provide some suggestions.

### DIFF
--- a/include/control_filters/exponential_filter.hpp
+++ b/include/control_filters/exponential_filter.hpp
@@ -87,7 +87,8 @@ bool ExponentialFilter<T>::configure()
                                                          this->param_prefix_);
     }
     catch (const std::exception & ex) {
-      RCLCPP_ERROR((*logger_), "Exponential filter cannot be configured: %s (type : %s)", ex.what(), typeid(ex).name());
+      RCLCPP_ERROR((*logger_),
+        "Exponential filter cannot be configured: %s (type : %s)", ex.what(), typeid(ex).name());
       parameter_handler_.reset();
       return false;
     }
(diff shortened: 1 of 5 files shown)


christophfroehlich (changes requested):
Thank you for your contribution!

Please fix the pre-commit issues (see the failing CI jobs). You can check that locally by `pre-commit run --all`, instructions for that you can find in the good first issue text or the contribution guidelines on control.ros.org

Regarding `catch (...)`: I suggest adding a similar log like in the [controller_manager](https://github.com/ros-controls/ros2_control/blob/53656e4636ba0472a5ee20d3a29f096ff50f6f5d/controller_manager/src/controller_manager.cpp#L552-L559)?

bmagyar (dismissed):
Ah this is much nicer, thank you!

saikishor (changes requested):
Changes look good to me. 

Just pre-commit fixes

Juliaj:
@christophfroehlich , thanks for the `pre-commit` pointer. My bad, I didn't run it with `--all`. Fix made. 

As for `catch(...)`, thanks for sharing an example. The `load_controller` takes in two the input parameters which are logged as part of the catch all log statement. For our use case, the method `configure()` takes no parameter. I made an update to log a general statement. Feedback is appreciated.

christophfroehlich (approved):
Thanks!

saikishor (approved):
Looks great.
Thanks for adding tests!

PR #264, opened 2025-01-11
Update include paths of GPL

christophfroehlich:
Needs [release 0.4.0](https://github.com/PickNikRobotics/generate_parameter_library/pull/232)
for reference, see https://github.com/PickNikRobotics/generate_parameter_library/pull/213


saikishor (dismissed):

saikishor:
I don't understand why ABI is breaking here

christophfroehlich:
> I don't understand why ABI is breaking here

ABI used main instead of testing, like we have in ros2_controllers for example

saikishor (approved):
I think it makes sense to use testing here

christophfroehlich:
@mergifyio backport humble

Issue #276, opened 2025-01-21
Understanding the Anti-Windup Action


ViktorCVS:
I'm creating this issue to gain a better understanding of the Anti-Windup Action in the PID controller code ([pid.cpp#L160C1-L173C4](https://github.com/ros-controls/control_toolbox/blob/ros2-master/src/pid.cpp#L160C1-L173C4)).

Currently, when antiwindup is set to true, the integral error is capped between i_max and i_min, which aligns with the expected anti-windup functionality by preventing the integral term from accumulating excessively. However, when antiwindup is set to false, the integral contribution to the output is also capped between i_max and i_min (which is also an anti-windup technique).

This raises a couple of questions. First, is it necessary to cap the integral contribution to the output when the package already has an antiwindup parameter that manages the integral error? Second, in unstable systems, systems with high-frequency behavior, or systems with noise, the proportional and derivative parts of the error might also exceed the actuator limits. Shouldn't the system limit the entire output (or the individual contributions) rather than just the integral contribution to ensure stability and prevent the proportional and derivative terms from causing actuator saturation (if this feature is necessary)?

As suggestions for a simple improvement, I would either remove the second block ([pid.cpp#L170C1-L173C4](https://github.com/ros-controls/control_toolbox/blob/ros2-master/src/pid.cpp#L170C1-L173C4)), implement it also in the proportional and/or derivative contributions, or implement it in the final output cmd_ ([pid.cpp#L179](https://github.com/ros-controls/control_toolbox/blob/ros2-master/src/pid.cpp#L179)).

However, it would still be unclear why there's a parameter that activates the Anti-Windup action whether you set it to true or not. A more robust improvement, **if this saturation is required**, is to expand it to the other contributions (and/or for the entire control output) and add new parameters to clarify which value is saturating what. I would like to work on this, but first, I thought it was a good idea to bring up this discussion to assess the necessity of this change.

christophfroehlich:
I agree with you, I stumbled upon this already once but haven't questioned it that much back then.

The part was added with this commit https://github.com/ros-controls/control_toolbox/commit/de6346b258e4edd5545e22a937dd58aeadd05c32, but there is no discussion about that I could find.

[...]

PR #294, opened 2025-03-07
[Pid] Save `i_term` instead of error integral

christophfroehlich:
The internal state should be the i_term to avoid jumps in the output while tuning the integral gain. Reverts #33 

This PR does not change the current behavior, only the field in the published message and the result of `get_current_pid_errors`

I got confused of the message definition, because it has some duplicates: 
https://github.com/ros-controls/control_msgs/pull/173
If yes, let's fix that in a new PR.


saikishor (commented):
In general, LGTM

I'll review antiwindup part again will approve it

Amronos (approved):
LGTM! The documentation though should be updated for the ``antiwindup`` parameter of the ``pid_controller``. As the documentation currently states that ``i_term_`` is always limited regardless of what the value of the ``antiwindup`` parameter is, which was the case before this PR.

src/pid.cpp
```diff
@@ -194,29 +194,21 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+    i_term_ += gains.i_gain_ * dt_s * p_error_;
   }
 
   // Calculate derivative contribution to command
   d_term = gains.d_gain_ * d_error_;
 
   // Compute the command
-  cmd_ = p_term + i_term + d_term;
+  // Limit i_term so that the limit is meaningful in the output
+  cmd_ = p_term +  std::clamp(i_term_, gains.i_min_, gains.i_max_) + d_term;
```
bijoua29, line 211: Not sure why i_term is being clamped again here when it was clamped above when antiwindup is true. Shouldn't it only be clamped when antiwindup is true

src/pid_ros.cpp
```diff
@@ -284,7 +284,7 @@ void PidROS::publish_pid_state(double cmd, double error, rclcpp::Duration dt)
       rt_state_pub_->msg_.error = error;
       rt_state_pub_->msg_.error_dot = d_error;
       rt_state_pub_->msg_.p_error = p_error;
-      rt_state_pub_->msg_.i_error = i_error;
+      rt_state_pub_->msg_.i_error = i_term;
```
bijoua29, line 287: I think you said you will change the message in a different PR since this is really weighted i-error.

src/pid.cpp
christophfroehlich: I left that to keep the current behavior, it was clamped even with antiwindup being true.
from the docstring:
```
antiwindup Antiwindup functionality. When set to true, limits
        the integral error to prevent windup; otherwise, constrains the
        integral contribution to the control output. i_max and
        i_min are applied in both scenarios.
```

this was already questioned in #276 and I didn't want to change this behavior in this PR.

PR #298, opened 2025-03-11
Add three new anti-windup techniques and a Saturation feature

ViktorCVS:
### **Overview**

This PR adds three new anti-windup techniques: back‑calculation, the conditioning technique, and conditional integration. It also adds a saturation feature for the PID output. New parameters have been introduced, and additional overloads have been implemented to ensure compatibility.

- Added three new anti-windup techniques
  - back-calculation
  - conditioning technique
  - conditional integration
- Added saturation feature to PID output

### **About compatibility**

The packages compile correctly and have passed the pre‑commit and colcon tests (packages with dependencies continue to show the same number of failures before and after my modifications). If the new parameters are not used, the package retains its old behavior.

### **About the older anti-windup technique**

My plan, either by the end of this PR or in a subsequent one, is to completely remove the older anti‑windup technique that has been used so far. This method, which is a form of conditional integration, has several disadvantages:
- If set incorrectly, it may cause a steady‑state error.
- If set incorrectly, it may not affect the system at all.
- Even if it is set between the steady‑state error limit and the value beyond which it has no effect, it is still difficult to find a configuration that improves the system as effectively as the other techniques.

Additionally, regardless of whether the 'antiwindup' parameter is set to true or false, the anti-windup technique is applied (using the same method with a different approach), so the user does not have the option to disable it.

### **About unit tests**

I've added 10 new unit tests for the new features and updated the existing ones to accommodate the new parameters.

### **Related PR's**

- ros-controls/ros2_controllers#1585
- ros-controls/control_msgs#178

### **Important notes**

These three techniques are common anti‑windup strategies used to mitigate the windup effect and are widely employed in control applications: back‑calculation [1], the conditioning technique [1,2], and conditional integration [1,3].

The default values for the tracking time constant are defined in [3,4] for back‑calculation and in [1] for the conditioning technique.

Both back‑calculation and the conditioning technique use forward Euler discretization; this may change before merging this PR.

### **Graphs**

I tested it on ros2_control_demos to better illustrate this feature and test it on simulation to valide the equations. The tests were conducted using a modified version of Example 1: [RRBot](https://github.com/ros-controls/ros2_control_demos/blob/master/example_1), which uses a PID controller instead of the default forward position controller. It was tested on Docker, Ubuntu Noble, and Jazzy. 

PID values: p = 4.0, i = 25.0, d = 0.5; u_max = 13, u_min = -13; and the tracking time constant was left at its default value.

unsat+sat+back

The standard response with a settling time (ts) of 5.2 seconds, the response affected by saturation, resulting in a settling time (ts_sat) of 8.6 seconds (+65.4% increase) and the response using the back-calculation technique, which improves performance with a settling time (ts_back) of 4.1 seconds (–21.2% decrease), even lower than the standard response.

back+ct+cin

back+ct+cin2

Those figures compares three anti-windup methods applied to the step response, a zoomed-in view of the step response is provided here to clearly distinguish between the three anti-windup strategies. They are all very similar due to the system and PID values, but they may vary significantly between applications.

cs_unsat+sat+back

The standard control output, the control output affected by saturation, with a recovery time from saturation of 6.8s and the control output using the back-calculation technique, with a recovery time from saturation of 2s (-70.6%).

cs_back+ct+cin

cs_back+ct+cin2

Those figures compares three control outputs using anti-windup methods, a zoomed-in view of the control output is provided here to clearly distinguish between the three anti-windup strategies. They are all very similar due to the system and PID values, but they may vary significantly between applications.

All the equations have been validated with these simulations, providing a feature with three techniques to address windup.

### **Final notes**

### **References**

[1] VISIOLI, A. Pratical PID Control. London: Springer-Verlag London Limited, 2006. 476 p.
[2] VRANCIC, D et al. Antiwindup, Bumpless, and Conditioned Transfer Techniques for PID Controllers. IEEE Control Systems Magazine, 1996. 16(4):48 - 57.
[3] BOHN, C.; ATHERTON, D. An analysis package comparing pid anti-windup strategies.
IEEE Control Systems Magazine, p. 34–40, 1995.
[4] ASTRöM, K.; HäGGLUND, T. PID Controllers: Theory, Design and Tuning. ISA Press.
Research Triangle Park, USA: Springer-Verlag London Limited, 1995. 343 p.


ViktorCVS:
@christophfroehlich, it appears that no reviewers have been assigned to this PR. Could you please help with that? If you have time, I'd appreciate it if you could also take a look at the changes.

src/pid.cpp
```diff
@@ -194,21 +241,50 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
-  if (gains.antiwindup_) {
+  if (gains.antiwindup_ && gains.antiwindup_strat_ == "none") {
     // Prevent i_term_ from climbing higher than permitted by i_max_/i_min_
     i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_,
       gains.i_min_, gains.i_max_);
-  } else {
+  } else if (!gains.antiwindup_ && gains.antiwindup_strat_ == "none") {
     i_term_ += gains.i_gain_ * dt_s * p_error_;
+    i_term_ = std::clamp(i_term_, gains.i_min_, gains.i_max_);
   }
```
christophfroehlich, line 255: isn't this the same?

control_toolbox.rolling.repos
```diff
@@ -1,7 +1,7 @@
 repositories:
   control_msgs:
     type: git
-    url: https://github.com/ros-controls/control_msgs.git
+    url: https://github.com/ViktorCVS/control_msgs.git
```
christophfroehlich, line 4:
```suggestion
    url: https://github.com/ros-controls/control_msgs.git
```
I added this just to run the CI successfully in the meantime

christophfroehlich (commented):
Thx for this thorough PR, but it will need some time to properly review it

src/pid.cpp
ViktorCVS: No, with just the else clause, if gains.antiwindup_strat_ isn’t "none", it will fall through to the i_term_ equation. However, in my approach, i_term_ should be zero during the first iteration, and then obtain a value after the output is calculated (due to the anti-causal nature of anti-windup techniques, which require both the saturated and unsaturated cmd values to compute i_term_), effectively influencing the cmd on the next iteration.

christophfroehlich: but isn't the content of if and else branch the same code in the end? And could just be simplified to `if (gains.antiwindup_strat_ == "none")`

ViktorCVS: I see now, you are right. It is interesting that this code passes the unit tests for the old anti-windup behavior, even though we can simplify the two old conditions into just one.

However, I'm now considering the readability of the code. If I use only 'gains.antiwindup_strat_ == "none"' as the condition in this if statement, we would visually suppress the anti-windup variable/parameter, and people might not understand where this variable is being used, since it exists and is referenced throughout the code. At the same time, it's odd to have such redundancy in the ros2_control code.

I'd rather keep it this way until I remove the feature, so the variable remains visible. What do you think? I could remove the entire old behavior and address this issue in this PR or the next one.

christophfroehlich: as you implemented it here, this is different from the current behavior?
on the master branch: if antiwindup is false, then the i_term will windup and only the integral action for cmd_ will be clamped. This is a total different dynamic behavior. Only if antiwindup is true, then also i_term will be limited.
To keep the current behavior, here only 
` if (gains.antiwindup_ && gains.antiwindup_strat_ == "none") `
is necessary?

ViktorCVS: as I implemented it here, is the same as the current behavior. All tests for this anti-windup feature have been passed (70 tests, all of them passed), so I believe that I don't change anything about it. I tried what you said, to simplify to just one condition, removing the else condition and all tests also have been passed:

```
// Calculate integral contribution to command
  if (gains.antiwindup_strat_ == "none") {
    // Prevent i_term_ from climbing higher than permitted by i_max_/i_min_
    i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_,
      gains.i_min_, gains.i_max_);
  }
```

image

For now I guess is worth questioning that if the current tests for this feature really are enough to show any difference. Besides this, even I can simplify, if I do this, I would suppress the anti-windup variable/parameter from the function, and people may not understand why this variable exists. What do you think?

christophfroehlich: I don't think that the current tests are good enough here (as you have proven). But I suggested 
` if (gains.antiwindup_ && gains.antiwindup_strat_ == "none")`
where I think that this has the same behavior than the current rolling version.

ViktorCVS: Nice! I see where I should make edits to maintain the same behavior as the rolling branch. I'll commit these changes tomorrow morning.

ViktorCVS: I didn't want to create i_error_ again, so I decided to put cmd_ on a conditional.

ViktorCVS: Hi @christophfroehlich, I've pushed some updates to the PR. When you have a moment, could you please take another look and continue the review? Thanks a lot for your time and help.

bmagyar:
@ViktorCVS could you please resolve the current conflicts?

ViktorCVS:
> @ViktorCVS could you please resolve the current conflicts?

done

christophfroehlich (changes requested):
Thanks a lot for the nice work including tests etc. Please fix the pre-commit errors, and only some minor comments in the code

christophfroehlich:
> My plan, either by the end of this PR or in a subsequent one, is to completely remove the older anti‑windup technique that has been used so far. This method, which is a form of conditional integration, has several disadvantages:
 
Please add a deprecation notice to the code, as well as a warning on `std::cout` if "none" is configured, for example    `"This option will be removed by the ROS 2 Kilted Kaiju release."`

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -182,6 +230,39 @@ class Pid
...
+        'conditioning_technique', 'conditional_integration', or 'none'. Note that
+        the 'back_calculation' and 'conditioning_technique' strategies use the
+        tracking_time_constant parameter to tune the anti-windup behavior. When a strategy
+        other than 'none' is selected, it will override the controller's default anti-windup behavior.
+   *
+   * \throws An std::invalid_argument exception is thrown if i_min > i_max or u_min > u_max
+   */
+  Pid(double p, double i, double d, double i_max, double i_min,
+    double u_max, double u_min, double trk_tc, bool saturation,
+    bool antiwindup, std::string antiwindup_strat);
```
christophfroehlich, line 264: thinking once again, maybe we shouldn't use std::string here in the base class, but an enum instead?
the pid_ros can have the string parameter, and maybe do the conversion manually there, or use a pattern like [here](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface%2Finclude%2Fhardware_interface%2Fhardware_info.hpp#L136-L191).

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -433,8 +605,9 @@ class Pid
   double p_error_last_; /** Save state for derivative state calculation. */
   double p_error_;      /** Error. */
   double d_error_;      /** Derivative of error. */
```
saikishor, line 607:
```suggestion
   double p_error_last_ = 0.0; /** Save state for derivative state calculation. */
   double p_error_ = 0.0 ;      /** Error. */
   double d_error_ = 0.0;      /** Derivative of error. */
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -433,8 +605,9 @@ class Pid
...
-  double i_term_;       /** Integrator state. */
+  double i_term_{0};    /** Integrator state. */
   double cmd_;          /** Command to send. */
+  double cmd_unsat_;    /** command without saturation. */
```
saikishor, line 610:
```suggestion
  double i_term_ = 0.0;    /** Integrator state. */
   double cmd_ = 0.0;          /** Command to send. */
  double cmd_unsat_ = 0.0;    /** command without saturation. */
```
Let's initialize all the member variables

control_toolbox/src/pid.cpp
```diff
@@ -225,23 +268,53 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+    // Limit cmd_ if saturation is enabled
+    cmd_ = std::clamp(cmd_unsat_, gains.u_min_, gains.u_max_);
+  } else {
+    cmd_ = cmd_unsat_;
+  }
+
+  if (gains.antiwindup_strat_ == "back_calculation" && gains.i_gain_ != 0) {
+    if (gains.trk_tc_ == 0.0 && gains.d_gain_ != 0.0) {
+      // Default value for tracking time constant for back calculation technique
+      gains.trk_tc_ = std::sqrt(gains.d_gain_/gains.i_gain_);
```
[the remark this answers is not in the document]
saikishor: I agree with @christophfroehlich

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -184,6 +220,37 @@ class PidROS
...
+   * \param p The proportional gain.
+   * \param i The integral gain.
+   * \param d The derivative gain.
+   * \param i_max The max integral windup.
+   * \param i_min The min integral windup.
+   * \param u_max Upper output clamp.
+   * \param u_min Lower output clamp.
+   * \param trk_tc Specifies the tracking time constant for the 'back_calculation'
+        and 'conditioning_technique' strategies. If set to 0.0 when one of these
+        strategies is selected, a recommended default value will be applied.
```
saikishor, line 234: What's the recommended value?

control_toolbox/src/pid_ros.cpp
```diff
@@ -20,18 +20,8 @@
...
-// BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-// POSSIBILITY OF SUCH DAMAGE.
-
-#include <algorithm>
-#include <cmath>
+// FOR A PARTICULAR git config pull.rebase true
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit 629337b.

control_toolbox/src/pid_ros.cpp
```diff
@@ -218,26 +243,38 @@ void PidROS::declare_param(const std::string & param_name, rclcpp::ParameterValu
...
+}
+
+void PidROS::initialize_from_args(double p, double i, double d, double i_max, double i_min,
+  double u_max, double u_min, double trk_tc, bool saturation, bool antiwindup,
+  std::string antiwindup_strat, bool save_i_term)
+{
+  if (i_min > i_max || u_min > u_max) {
+    RCLCPP_ERROR(node_logging_->get_logger(),
+    "received i_min > i_max or u_min > u_max, skip new gains");
+  } else {
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit a683251. I've separated them with if else.

control_toolbox/src/pid_ros.cpp
```diff
@@ -276,20 +313,31 @@ Pid::Gains PidROS::get_gains() { return pid_.get_gains(); }
...
-  else
-  {
+  set_gains(p, i, d, i_max, i_min, 0.0, 0.0, 0.0, false, antiwindup, "none");
+}
+
+void PidROS::set_gains(double p, double i, double d, double i_max, double i_min, double u_max,
+  double u_min, double trk_tc, bool saturation, bool antiwindup, std::string antiwindup_strat)
+{
+  if (i_min > i_max || u_min > u_max) {
+    RCLCPP_ERROR(node_logging_->get_logger(),
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit a683251. I've separated them with if else.

control_toolbox/src/pid_ros.cpp
```diff
@@ -340,28 +388,31 @@ void PidROS::print_values()
...
 {
-  if (gains.i_min_ > gains.i_max_)
-  {
-    RCLCPP_ERROR(node_logging_->get_logger(), "received i_min > i_max, skip new gains");
-  }
-  else
-  {
-    pid_.set_gains(gains);
+  if (gains.i_min_ > gains.i_max_ || gains.u_min_ > gains.u_max_) {
+    RCLCPP_ERROR(node_logging_->get_logger(),
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit a683251. I've separated them with if else.

control_toolbox/src/pid_ros.cpp
```diff
@@ -421,13 +485,11 @@ void PidROS::set_parameter_event_callback()
...
       /// @note don't call set_gains() from inside a callback
-      if (gains.i_min_ > gains.i_max_)
-      {
-        RCLCPP_ERROR(node_logging_->get_logger(), "received i_min > i_max, skip new gains");
-      }
-      else
-      {
-        pid_.set_gains(gains);
+      if (gains.i_min_ > gains.i_max_ || gains.u_min_ > gains.u_max_) {
+        RCLCPP_ERROR(node_logging_->get_logger(),
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit a683251. I've separated them with if else.

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -137,28 +164,49 @@ class Pid
...
+    double i_min_;                 /**< Minimum allowable integral term. */
+    double u_max_;                 /**< Maximum allowable output. */
+    double u_min_;                 /**< Minimum allowable output. */
+    double trk_tc_;                /**< Tracking time constant. */
+    bool saturation_;              /**< Saturation. */
+    bool antiwindup_;              /**< Anti-windup. */
+    std::string antiwindup_strat_; /**< Anti-windup strategy. */
   };
 
   /*!
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit f561b70

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -182,6 +230,39 @@ class Pid
     double p = 0.0, double i = 0.0, double d = 0.0, double i_max = 0.0, double i_min = -0.0,
     bool antiwindup = false);
 
+  /*!
+   * \brief Constructor, zeros out Pid values when created and
+   *        initialize Pid-gains and integral term limits.
+   *        Does not initialize dynamic reconfigure for PID gains
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit f561b70

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -212,6 +293,37 @@ class Pid
   void initialize(
     double p, double i, double d, double i_max, double i_min, bool antiwindup = false);
 
+  /*!
+   * \brief Optional constructor for passing in values
```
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit f561b70

control_toolbox/src/pid.cpp
```diff
@@ -225,23 +268,53 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
     i_term_ += gains.i_gain_ * dt_s * p_error_;
   }
 
-  // Calculate derivative contribution to command
-  d_term = gains.d_gain_ * d_error_;
-
   // Compute the command
   // Limit i_term so that the limit is meaningful in the output
-  cmd_ = p_term + std::clamp(i_term_, gains.i_min_, gains.i_max_) + d_term;
+  if (!gains.antiwindup_ && gains.antiwindup_strat_ == "none") {
```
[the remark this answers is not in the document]
ViktorCVS: Indeed! Fixed in commit 8233898.

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -184,6 +220,37 @@ class PidROS
...
+   * \brief Initialize the PID controller and set the parameters
+   * \param p The proportional gain.
+   * \param i The integral gain.
+   * \param d The derivative gain.
+   * \param i_max The max integral windup.
+   * \param i_min The min integral windup.
+   * \param u_max Upper output clamp.
+   * \param u_min Lower output clamp.
+   * \param trk_tc Specifies the tracking time constant for the 'back_calculation'
+        and 'conditioning_technique' strategies. If set to 0.0 when one of these
```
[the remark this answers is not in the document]
ViktorCVS: Yes! Everywhere that trk_tc needs to be initialized, I set it to 0.0. I also set this variable’s default value to 0.0 in ros2_controllers.

control_toolbox/src/pid_ros.cpp
```diff
@@ -182,27 +172,62 @@ bool PidROS::get_double_param(const std::string & param_name, double & value)
   }
 }
 
+bool PidROS::get_string_param(const std::string & param_name, std::string & value)
```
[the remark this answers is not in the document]
ViktorCVS: I’ve used generate_parameter_library in my recent projects, but I’m not sure if other approaches might be better.

control_toolbox/src/pid.cpp
```diff
@@ -225,23 +268,53 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+
+  if (gains.saturation_ == true) {
+    // Limit cmd_ if saturation is enabled
+    cmd_ = std::clamp(cmd_unsat_, gains.u_min_, gains.u_max_);
+  } else {
+    cmd_ = cmd_unsat_;
+  }
+
+  if (gains.antiwindup_strat_ == "back_calculation" && gains.i_gain_ != 0) {
+    if (gains.trk_tc_ == 0.0 && gains.d_gain_ != 0.0) {
```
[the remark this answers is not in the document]
ViktorCVS: Nice! I added "inline" tag and putted it in pid.hpp. I also created a "is_not_zero(x)" to adress !=0.0 instead of use !is_zero(x). If it isn't ok, I can keep only is_zero. Adressed in commit 4c1fc9b.

control_toolbox/test/pid_tests.cpp
```diff
@@ -209,15 +503,28 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
   EXPECT_EQ(i_max, g1.i_max_);
   EXPECT_EQ(i_min, g1.i_min_);
+  EXPECT_EQ(u_max, g1.u_max_);
+  EXPECT_EQ(u_min, g1.u_min_);
+  EXPECT_EQ(trk_tc, g1.trk_tc_);
+  EXPECT_EQ(saturation, g1.saturation_);
   EXPECT_EQ(antiwindup, g1.antiwindup_);
+  EXPECT_EQ(antiwindup_strat, g1.antiwindup_strat_);
 
   // \todo test initParam() -------------------------------------------------
```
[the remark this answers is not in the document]
ViktorCVS: Fixed in commit e1e6d05.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Fixed in commit e1b7413.

control_toolbox/src/pid_ros.cpp
```diff
@@ -421,13 +485,11 @@ void PidROS::set_parameter_event_callback()
     if (changed)
     {
       /// @note don't call set_gains() from inside a callback
```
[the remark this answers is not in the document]
christophfroehlich: yes yes I know that the comment was there before, but I maybe never have read it. And as you propose a change to the callback method, we can also think about deleting the comment because it does not make sense (any more?)?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -125,53 +144,131 @@ class Pid
...
+    double i_gain_;
+    double d_gain_;
+    double i_max_;
+    double i_min_;
+    double u_max_;
+    double u_min_;
+    double trk_tc_;
+    bool saturation_;
+    bool antiwindup_;
+    std::string antiwindup_strat_;
```
christophfroehlich, line 259: I think it make sense to re-add the code comments deleted in https://github.com/ros-controls/control_toolbox/commit/f561b704ceb9bbb54a0a68edca5ccc12a953b71e, what do you think?

control_toolbox/src/pid_ros.cpp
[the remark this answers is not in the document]
christophfroehlich: https://github.com/ros-controls/control_toolbox/issues/350

control_toolbox/src/pid.cpp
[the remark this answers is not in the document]
christophfroehlich: Personally I prefer `!is_zero`, but as you like

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -426,15 +644,14 @@ class Pid
   }
 
 protected:
-  // Store the PID gains in a realtime buffer to allow dynamic reconfigure to update it without
-  // blocking the realtime update loop
   realtime_tools::RealtimeBuffer<Gains> gains_buffer_;
```
christophfroehlich, line 647: this comment still applies?
```suggestion
  // Store the PID gains in a realtime buffer to allow dynamic reconfigure to update it without
  // blocking the realtime update loop
  realtime_tools::RealtimeBuffer<Gains> gains_buffer_;
```

control_toolbox/include/control_toolbox/pid_ros.hpp
christophfroehlich: I also realized that later in the review

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -184,6 +220,38 @@ class PidROS
...
+        tracking_time_constant parameter to tune the anti-windup behavior. When a strategy
+        other than 'none' is selected, it will override the controller's default
+        anti-windup behavior.
+   *
+   * \note New gains are not applied if i_min > i_max or if u_min_ > u_max_.
+   */
+  void set_gains(
+    double p, double i, double d, double i_max, double i_min, double u_max, double u_min,
+    double trk_tc, bool saturation = false, bool antiwindup = false,
+    std::string antiwindup_strat = "none");
```
christophfroehlich, line 253: that's what I meant with default value
```suggestion
  void set_gains(
    double p, double i, double d, double i_max, double i_min, double u_max, double u_min,
    double trk_tc = 0.0, bool saturation = false, bool antiwindup = false,
    std::string antiwindup_strat = "none");
```

christophfroehlich (changes requested):
For the future: Please don't force push to PRs because it makes it harder for reviewers to check the changes since the last review ;) The history does not have to be linear, because we squash that anyways.

We get closer to the finish line, but some of our comments haven't been addressed yet.

control_toolbox/src/pid_ros.cpp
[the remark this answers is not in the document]
ViktorCVS: I agree! Removed in commit d6e00b4.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: I've deleted it at first because I thought dynamic reconfigure is ROS1 related. Added in commit d9a6f2e.

control_toolbox/src/pid.cpp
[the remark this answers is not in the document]
ViktorCVS: I agree. Change it to use only is_zero() in commit 5513ff0.

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: yes, the dynamic_reconfigure was a package in ROS 1 to specifically enable this, but the term describes it still well for ROS 2 ;)

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: I had some problems on this, but change it in commit b7c8866, 9826771

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -34,12 +34,84 @@
...
+    {
+      value_ = CONDITIONING_TECHNIQUE;
+    }
+    else if (s == "conditional_integration")
+    {
+      value_ = CONDITIONAL_INTEGRATION;
+    }
+    else
+    {
+      // Qualquer valor desconhecido agora cai em NONE
```
christophfroehlich, line 76:
```suggestion
```
not everyone will understand Portuguese ;)

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -34,12 +34,84 @@
...
+  enum Value : int8_t
+  {
+    NONE = 0,
+    BACK_CALCULATION,
+    CONDITIONING_TECHNIQUE,
+    CONDITIONAL_INTEGRATION
+  };
+
+  constexpr AntiwindupStrategy() : value_(NONE) {}
+  explicit constexpr AntiwindupStrategy(Value v) : value_(v) {}
```
christophfroehlich, line 58:
```suggestion
  constexpr AntiwindupStrategy(Value v) : value_(v) {}  // NOLINT(runtime/explicit)
```

And then the constructor can be omitted and this just works
```
  Pid pid(
    0.0, 1.0, 0.0, 0.0, 0.0, 5.0, -5.0, 1.0, true, false,
    AntiwindupStrategy::BACK_CALCULATION);
```
What do you think? This can be simplified in the tests then

christophfroehlich (changes requested):
The tests are failing now, can you have a look please?

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: That helped a lot. I was unsure whether I could use NOLINT—thanks! Commit e89ac18 addresses this.

control_toolbox/src/pid.cpp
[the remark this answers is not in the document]
ViktorCVS: Thanks! Fixed in commit 64ce2d6.

christophfroehlich:
> For the future: Please don't force push to PRs because it makes it harder for reviewers to check the changes since the last review ;) The history does not have to be linear, because we squash that anyways.

ViktorCVS:
> > For the future: Please don't force push to PRs because it makes it harder for reviewers to check the changes since the last review ;) The history does not have to be linear, because we squash that anyways.

Sorry about that. I tried pushing without force, but Git wouldn’t accept it. I’ll try other methods from now on.

ViktorCVS:
> why you rebased and force-pushed again? the diff is strange now because the last commit on master is included here?

There’s a warning that this branch is out of date with the base branch, so I pulled the latest changes and pushed. This time I didn’t use a force push, so it’s strange that you’re still seeing this message.

Whenever the master branch is updated, I can’t push without force. Since you asked me to avoid force pushes, I merged master into my branch before pushing.

christophfroehlich:
alright, I see what I can do to fix it and do the final review round ;)

christophfroehlich:
I merged them already in, now the diff view looks fixed again.

control_toolbox/src/pid.cpp
```diff
@@ -212,36 +290,71 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+
+  if (gains.antiwindup_strat_ == AntiwindupStrategy::BACK_CALCULATION && !is_zero(gains.i_gain_))
+  {
+    i_term_ += dt_s * (gains.i_gain_ * error + 1 / gains.trk_tc_ * (cmd_ - cmd_unsat_));
+  }
+  else if (
+    gains.antiwindup_strat_ == AntiwindupStrategy::CONDITIONING_TECHNIQUE &&
+    !is_zero(gains.i_gain_))
+  {
+    i_term_ += dt_s * gains.i_gain_ * (error + 1 / gains.trk_tc_ * (cmd_ - cmd_unsat_));
```
christophfroehlich, line 349: I don't have access to [2], but in [1] the conditioning technique is described differently (Fig 3.4). 

Is there a real benefit in adding all of the three? Or should we skip the CONDITIONING_TECHNIQUE as it is the least known/documented one? I think it leaves enough choices and tuning possibilities for the user.

control_toolbox/src/pid.cpp
```diff
@@ -129,21 +170,58 @@ void Pid::set_gains(double p, double i, double d, double i_max, double i_min, bo
...
   else
   {
+    Gains gains = gains_in;
+
+    if (gains.antiwindup_strat_ == AntiwindupStrategy::BACK_CALCULATION)
+    {
+      if (is_zero(gains.trk_tc_) && !is_zero(gains.d_gain_))
+      {
+        // Default value for tracking time constant for back calculation technique
+        gains.trk_tc_ = std::sqrt(gains.d_gain_ / gains.i_gain_);
```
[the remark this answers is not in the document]
ViktorCVS: yes. in [1] p.39 eq. 3.3, he defines $T_{t}=\sqrt{T_{i}T_{d}}$. And since control_toolbox uses the parallel form for PID (kp,ki,kd) a conversion has to be made for both parameters. $T_{i}=\frac{K_{p}}{K_{i}}$ and $T_{d}=\frac{K_{d}}{K_{p}}$, so:
$T_{t}=\sqrt{\frac{K_{d}K_{p}}{K_{i}K_{p}}}=\sqrt{\frac{K_{d}}{K_{i}}}$. There's also a work in portuguese "Estudo Comparativo de Estratégias Anti-Windup
Aplicadas a Sistemas de Controle/Comparative Study of Anti-Windup Strategies Applied to Control Systems) that list the equation this way.

control_toolbox/src/pid.cpp
ViktorCVS: I simplified the material from source [2], but I have researched the CONDITIONING_TECHNIQUE and noticed some discrepancies between sources. In some of them, BACK_CALCULATION and CONDITIONING_TECHNIQUE can have the same effect if we tune Tt in a specific way. Therefore, I agree to remove CONDITIONING_TECHNIQUE. I’ll do this as soon as possible. I’ll also replace [2] with an equivalent but accessible article.

christophfroehlich: great thx, let's push this to be merged soon. documentation can be added afterwards, because we don't need to release it

ViktorCVS: solved in commit 5b3b630 and 9610d41. I also don't know why it made two commits

ViktorCVS:
> I merged them already in, now the diff view looks fixed again.

Can you take a look again? My Git didn't work without push with force, so I push the next commit with /f.

control_toolbox/src/pid.cpp
```diff
@@ -212,36 +290,71 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+  }
+  else if (
+    gains.antiwindup_strat_ == AntiwindupStrategy::CONDITIONING_TECHNIQUE &&
+    !is_zero(gains.i_gain_))
+  {
+    i_term_ += dt_s * gains.i_gain_ * (error + 1 / gains.trk_tc_ * (cmd_ - cmd_unsat_));
+  }
+  else if (gains.antiwindup_strat_ == AntiwindupStrategy::CONDITIONAL_INTEGRATION)
+  {
+    if (!(cmd_unsat_ != cmd_ && error * cmd_unsat_ > 0))
```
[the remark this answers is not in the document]
ViktorCVS: True! Fixed in a61cf71

control_toolbox/src/pid.cpp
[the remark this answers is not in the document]
christophfroehlich: you are right, thanks for checking again.

christophfroehlich (changes requested):
Oh, I see that the downstream PidController tests fail now. Can you have a look please and update https://github.com/ros-controls/ros2_controllers/pull/1585?

christophfroehlich:
> Fixed
Thanks. I see now that `antiwindup` parameter is not used any more if a strategy other than NONE is configured. Shouldn't we add this flag there also?

control_toolbox/src/pid_ros.cpp
```diff
@@ -275,21 +349,45 @@ double PidROS::compute_command(double error, double error_dot, const rclcpp::Dur
...
+  else if (u_min > u_max)
+  {
+    RCLCPP_ERROR(node_logging_->get_logger(), "received u_min > u_max, skip new gains");
+  }
   else
   {
+    if (antiwindup_strat == AntiwindupStrategy::NONE)
+    {
+      std::cout << "Old anti-windup technique is deprecated. "
+                   "This option will be removed by the ROS 2 Kilted Kaiju release."
```
saikishor, line 373:
```suggestion
                   "This option will be removed after the ROS 2 Kilted Kaiju releases."
```
Should it be after? @christophfroehlich

saikishor (approved):
@ViktorCVS Very good work

I've only one question, what will the bool antiwindup parameter be used for from now on?

ViktorCVS:
> @ViktorCVS Very good work
> 
> I've only one question, what will the bool antiwindup parameter be used for from now on?

Thanks! Until I remove it, it will still maintain the old behavior. This PR just adds more options. I’ll open a follow-up PR today to remove it.

control_toolbox/src/pid_ros.cpp
saikishor: I would opt for backporting it as it comes with very nice features + easy transition

christophfroehlich: it does not break API but ABI, and that's what I try to avoid :/

ViktorCVS:
> > Fixed
> 
> Thanks. I see now that `antiwindup` parameter is not used any more if a strategy other than NONE is configured. Shouldn't we add this flag there also?

In ros2_controllers? I agree. I’ll update the anti-windup parameter description accordingly.

ViktorCVS:
> Thanks again!
> 
> Could you please submit two follow-up PRs:
> 
> * One which updates the documentation, basically the summary of your initial message here. But you don't need to document the old behavior, just the result after the deprecation cleanup. Here you can add the migration notes.
> * Another one cleaning up the deprecated code. You can draft it for now, and we will merge after the next sync to kilted/rolling.

Should I wait for the merge?

control_toolbox/src/pid.cpp
```diff
@@ -212,36 +282,65 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+  {
+    i_term_ += dt_s * (gains.i_gain_ * error + 1 / gains.trk_tc_ * (cmd_ - cmd_unsat_));
+  }
+  else if (gains.antiwindup_strat_ == AntiwindupStrategy::CONDITIONAL_INTEGRATION)
+  {
+    if (!(!iszero(cmd_unsat_ - cmd_) && error * cmd_unsat_ > 0))
+    {
+      i_term_ += dt_s * gains.i_gain_ * error;
+    }
+  }
```
christophfroehlich, line 343: I mean using `gains.antiwindup_` here? How else can it be deactivated? Or is the plan to leave the NONE for this case?

ViktorCVS: Yes! if the user doesn’t want any anti-windup effect on the PID, they should set the value to “none"

ViktorCVS: Removing the anti-windup parameter from the constructor that accepts multiple arguments would render it inaccessible to users. I suggest we preserve the existing functionality for this PR. What do you think?

christophfroehlich: Let's deprecate all methods (constructors, initialize, set_gains,.) having `bool antiwindup` but no `AntiwindupStrategy`, and remove the bool from the overloaded variants with `AntiwindupStrategy`.

This should not break current behaviors, but helps us in the cleanup afterwards. What do you think @ViktorCVS?

ViktorCVS: Sorry for the delay. I agree; this will make it clearer what will be removed and help clean things up.

saikishor: What about moving this `AntiwindupStrategy`  as a non-optional arg of the methods, this should give us some flexibility. I think it makes sense to be explicitly set.

christophfroehlich: the other overloads are here just for backward compatibility, whenever a new parameter was added. do you propose to deprecate+remove them all?

ViktorCVS: @christophfroehlich I’ve been reviewing the code, and it seems that if we drop the bool antiwindup parameter from the full overload, we’ll lose any way to pass that flag while it still exists. The wrapper that takes only the bool currently forwards its value to the full overload. If the full overload no longer accepts that argument, there’s nowhere for it to go. Am I overlooking something?

christophfroehlich: why not adding the future full-overload, and deprecating the current full-overload with the bool? You can use gcc pragmas to avoid deprecation warnings within our own code:

```
#pragma GCC diagnostic push
<deprecated code..>
#pragma GCC diagnostic pop
```

ViktorCVS: Done in commit 257fb57

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -260,6 +262,7 @@ class Pid
         than 'none' is selected, it will override the controller's default anti-windup behavior.
    *
    */
+    [[deprecated("Use constructor with AntiwindupStrategy only.")]]
     Gains(
```
christophfroehlich, line 266: do we need this overload?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -351,10 +389,33 @@ class Pid
    *
    * \throws An std::invalid_argument exception is thrown if i_min > i_max or u_min > u_max
    */
+  [[deprecated("Use constructor with AntiwindupStrategy only.")]]
   Pid(
```
christophfroehlich, line 393: do we need this overload?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -409,10 +471,33 @@ class Pid
    *
    * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
    */
+  [[deprecated("Use initialize with AntiwindupStrategy only.")]]
   void initialize(
```
christophfroehlich, line 475: do we need this overload?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -478,11 +564,31 @@ class Pid
         tracking_time_constant parameter to tune the anti-windup behavior. When a strategy other
         than 'none' is selected, it will override the controller's default anti-windup behavior.
    */
+  [[deprecated("Use get_gains overload with AntiwindupStrategy only.")]]
   void get_gains(
```
christophfroehlich, line 568: do we need this overload?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -529,11 +636,34 @@ class Pid
    *
    * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
    */
+  [[deprecated("Use set_gains with AntiwindupStrategy only.")]]
   void set_gains(
```
christophfroehlich, line 640: do we need this overload?

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -150,11 +152,35 @@ class PidROS
    *
    * \note New gains are not applied if i_min_ > i_max_ or if u_min_ > u_max_.
    */
+  [[deprecated("Use initialize_from_args with AntiwindupStrategy only.")]]
   void initialize_from_args(
```
christophfroehlich, line 156: do we need this overload?

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -252,11 +279,34 @@ class PidROS
    *
    * \note New gains are not applied if i_min > i_max or if u_min_ > u_max_.
    */
+  [[deprecated("Use set_gains with AntiwindupStrategy only.")]]
   void set_gains(
```
christophfroehlich, line 283: do we need this overload?

christophfroehlich (changes requested):
Do we need the overloads with the boolean AND the new AntiwindupStrategy?

control_toolbox/src/pid.cpp
christophfroehlich: Thanks!

christophfroehlich: maybe I have described it wrong, the "intermediate" full-overload with bool is not needed as we haven't merged that yet, right?

ViktorCVS: I’m using it to get the bool antiwindup input from the user and use it to compute cmd_. If I remove the intermediate full‐overload, this parameter won’t be passed to compute_command and we’ll break the current behavior since it forwards its value

christophfroehlich: ok we could reimplement the methods using the correct value. but nevermind, we have to break API anyways for the cleanup.

saikishor: > Done in commit [257fb57](https://github.com/ros-controls/control_toolbox/commit/257fb57c7c4c554fdef644589001c5df523435f0)

The changes look good to me. Great work

> I’m using it to get the bool antiwindup input from the user and use it to compute cmd_. If I remove the intermediate full‐overload, this parameter won’t be passed to compute_command and we’ll break the current behavior since it forwards its value.

You mean this part of the code

```
 if (gains.antiwindup_ && gains.antiwindup_strat_ == AntiwindupStrategy::NONE)
  {
    // Prevent i_term_ from climbing higher than permitted by i_max_/i_min_
    i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_, gains.i_min_, gains.i_max_);
  }
  else if (!gains.antiwindup_ && gains.antiwindup_strat_ == AntiwindupStrategy::NONE)
  {
    i_term_ += gains.i_gain_ * dt_s * p_error_;
  }
```
I think it would be nice to remove the `antiwindup = false` in the newly introduced constructor

Can't we do something like, when we have the antiwindup = false, the default the strategy to None? this way you should be able to have it backward compatible right?
```

   double p, double i, double d, double i_max, double i_min, double u_max, double u_min,
    double trk_tc = 0.0, bool saturation = false, bool antiwindup = false,
    AntiwindupStrategy antiwindup_strat = AntiwindupStrategy::NONE);
```

saikishor: I'm sorry that I'm not active in most discussions, but I've been reading the notifications etc

christophfroehlich: No because I think we have the two different branches with AntiwindupStrategy::NONE for antiwindup true and false?

ViktorCVS: Yes, this is the part I’m referring to, but unfortunately we can’t disable bool antiwindup by setting it to false, since AntiwindupStrategy::NONE is valid whether bool antiwindup is false or true

saikishor: I just took a look at the code. Does it make sense to have 2 new Antiwindup strategies like `::CLAMP_I_TERM` and `::CLAMP_CMD_UNSATURATED` or something like that. So, based on `antiwindup` being true or not, we can internally set one of the other.

Any  opinion?

ViktorCVS: I thought about that and it seems good, but I hesitated because I didn’t want to modify the new feature during the cleanup. Do you think the code would be more organized/clean this way?

saikishor: I believe it is. We can merge it as it is now and work on that on a different PR before releasing it

christophfroehlich (approved):

christophfroehlich:
@ViktorCVS I came to the conlusion with Sai that we merge this and discuss about the saturation in a follow-up PR. Could you please prepare the documentation update first? The deprecation removal has more time, at least until we release the package.

PR #326, opened 2025-04-02
Don't return cmd if called with dt=0 or garbage

### LINKED ISSUE #31: pid::updateCommand inconsistency with dt=0 or errors
Line 342 of pid.cpp

If an error or dt=0, occurs you return zero but don't update the stored command.
I saw this causing issues in simulation

I suggest the following:

``` c++
  if (dt == ros::Duration(0.0))
  {
      return cmd_;
  }
  else if (std::isnan(error) || std::isinf(error) || std::isnan(error_dot) || std::isinf(error_dot))
  {
    i_term_ = 0.0;
    cmd_=0.0;
    return 0.0;
  }
```

I haven't pushed change to our fork yet, so just putting this up for discussion.

christophfroehlich:
Fixes #31 

- return the same command if update is called with dt=0
- return NaN if called with any error being NaN, but not change internal states for fast recovery afterwards.

### DIFF
--- a/control_toolbox/src/pid.cpp
+++ b/control_toolbox/src/pid.cpp
@@ -39,6 +39,7 @@
 #include <algorithm>
 #include <cmath>
 #include <iostream>
+#include <limits>
 #include <stdexcept>
 #include <utility>
 
@@ -143,7 +144,7 @@ double Pid::compute_command(double error, const double & dt_s)
 {
   if (dt_s < 0.0) {
     throw std::invalid_argument("Pid is called with negative dt");
-  } else if (dt_s <= 0.0) {
+  } else if (dt_s <=  std::numeric_limits<float>::epsilon()) {
     // don't update anything
     return cmd_;
   }
@@ -189,7 +190,7 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
 {
   if (dt_s < 0.0) {
     throw std::invalid_argument("Pid is called with negative dt");
-  } else if (dt_s <= 0.0) {
+  } else if (dt_s <=  std::numeric_limits<float>::epsilon()) {
     // don't update anything
     return cmd_;
   }
--- a/control_toolbox/src/pid.cpp
+++ b/control_toolbox/src/pid.cpp
@@ -141,8 +141,16 @@ void Pid::set_gains(const Gains & gains)
 
 double Pid::compute_command(double error, const double & dt_s)
 {
-  if (dt_s <= 0.0 || !std::isfinite(error)) {
-    return 0.0;
+  if (dt_s < 0.0) {
+    throw std::invalid_argument("Pid is called with negative dt");
+  } else if (dt_s <= 0.0) {
+    // don't update anything
+    return cmd_;
+  }
+
+  // reset controller
+  if (!std::isfinite(error)) {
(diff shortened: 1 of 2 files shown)


src/pid.cpp
```diff
@@ -179,16 +179,20 @@ double Pid::compute_command(
 
 double Pid::compute_command(double error, double error_dot, const double & dt_s)
 {
+  // don't update anything
+  if (dt_s <= 0.0) {
+    return cmd_;
+  }
+
```
saikishor, line 186: I would raise an exception if the dt_s is negative 

And when it is zero, cmd

src/pid.cpp
```diff
@@ -179,16 +179,20 @@ double Pid::compute_command(
...
 
   double p_term, d_term;
   p_error_ = error;  // this is error = target - state
   d_error_ = error_dot;
 
-  if (
-    dt_s <= 0.0 || !std::isfinite(error) || !std::isfinite(error_dot)) {
-    return 0.0;
+  if (!std::isfinite(error) || !std::isfinite(error_dot)) {
+    return cmd_ = i_term_ = 0.0;
```
saikishor, line 195: Do you need to do this?

Because iterm is only set later

christophfroehlich: If we reset the cmd here, I agree with
https://github.com/ros-controls/control_toolbox/issues/31#issuecomment-59842858

we can also set both to NaN and just proceed. But the PID can't recover anyways then

src/pid.cpp
saikishor: Because negative time should never happen in a monotonic slewless clock, and atleast with an exception the user will know that there is something wrong with his clock

src/pid.cpp
saikishor: I'm not sure at this point. If it's not finite, i personally prefer the output to be NaN, instead of Zero. Zero is still a valid value

src/pid.cpp
christophfroehlich: done

src/pid.cpp
saikishor: What do you think of this?

christophfroehlich: so you propose return NaN and reset the internal states, and it recovers the next time having a valid input?

control_toolbox/src/pid.cpp
```diff
@@ -141,8 +141,16 @@ void Pid::set_gains(const Gains & gains)
 
 double Pid::compute_command(double error, const double & dt_s)
 {
-  if (dt_s <= 0.0 || !std::isfinite(error)) {
-    return 0.0;
+  if (dt_s < 0.0) {
+    throw std::invalid_argument("Pid is called with negative dt");
+  } else if (dt_s <= 0.0) {
```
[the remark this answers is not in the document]
christophfroehlich: is this safer than comparing with `<=0`? I really don't know

src/pid.cpp
saikishor: I saw only return NaN and not reset internal states, as this can happen when one of the reading is NaN and we don't want the i_term to reset and it takes a time for the system to stabilize

saikishor (approved):
LGTM

christophfroehlich:
@mergifyio backport mergify/bp/jazzy/pr-400

PR #327, opened 2025-04-02
Fix clang-format

christophfroehlich:
pre-commit didn't report such errors in my local devcontainer, so I wondered why: This repository is missing a .clang-format file, and the fallback-style for clang-format is set to 'None' -> no format at all.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,15 @@
+---
+Language:        Cpp
+BasedOnStyle:  Google
+
+ColumnLimit: 100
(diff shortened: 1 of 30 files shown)


saikishor (approved):
Looks good to me
 

I haven't taken look at all the changes as they are formatting changes. The .clang_format file looks good

christophfroehlich:
I promise, these all are only changes from pre-commit, and we use the same format settings in all over our repos 😉

PR #339, opened 2025-04-18
Add Kilted Kaiju

christophfroehlich:
and branch ros2_control/ros2_controllers for jazzy


saikishor (approved):
LGTM.

Just looking at the logs. It seems we don't have the image "ghcr.io/ros-controls/ros:kilted-ubuntu" and "ros:kilted" images yet

christophfroehlich:
yes, only testing because there is no no-testing release yet

PR #340, opened 2025-04-22
Add std::vector<double> option to the low pass filter

MarqRazz:
This PR adds the option to create a `LowPassFilter` object with type `std::vector<double>` (seeing that most system interfaces are this type :smile: )

The velocity data from my robot is noisy so I needed to clean it up
image
to
image

### DIFF
--- a/control_toolbox/include/control_toolbox/low_pass_filter.hpp
+++ b/control_toolbox/include/control_toolbox/low_pass_filter.hpp
@@ -124,6 +124,8 @@ private:
   double b1_; /** feedforward coefficient. */
   /** internal data storage (double). */
   double filtered_value, filtered_old_value, old_value;
+  /** internal data storage (std::vector<double>). */
+  std::vector<double> vector_filtered_value, vector_filtered_old_value, vector_old_value;
   /** internal data storage (wrench). */
   Eigen::Matrix<double, 6, 1> msg_filtered, msg_filtered_old, msg_old;
   bool configured_ = false;
@@ -183,6 +185,35 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
 
   // copy the header
   data_out.header = data_in.header;
+  return true;
+}
+
+template <>
+inline bool LowPassFilter<std::vector<double>>::update(
+  const std::vector<double> & data_in, std::vector<double> & data_out)
+{
+  if (!configured_)
+  {
+    throw std::runtime_error("Filter is not configured");
+  }
+  // If this is the first call to update initialize the filter at the current state
+  // so that we dont apply an impulse to the data.
+  // This also sets the size of the member variables to match the input data.
+  if (vector_filtered_value.empty())
+  {
+    vector_filtered_value = data_in;
+    vector_filtered_old_value = data_in;
+    vector_old_value = data_in;
+  }
+
+  // Filter each value in the vector
+  for (std::size_t i = 0; i < data_in.size(); i++)
+  {
+    data_out[i] = b1_ * vector_old_value[i] + a1_ * vector_filtered_old_value[i];
+    vector_filtered_old_value[i] = data_out[i];
+    vector_old_value[i] = data_in[i];
+  }
+
   return true;
 }
 
--- a/control_toolbox/include/control_toolbox/low_pass_filter.hpp
+++ b/control_toolbox/include/control_toolbox/low_pass_filter.hpp
@@ -230,7 +230,7 @@ inline bool LowPassFilter<std::vector<double>>::update(
   {
     data_out[i] = b1_ * old_value[i] + a1_ * filtered_old_value[i];
     filtered_old_value[i] = data_out[i];
-    old_value[i] = data_in[i];
+    if (std::isfinite(data_in[i])) old_value[i] = data_in[i];
   }
 
   return true;
@@ -255,7 +255,7 @@ bool LowPassFilter<T>::update(const T & data_in, T & data_out)
   // Filter
   data_out = b1_ * old_value + a1_ * filtered_old_value;
   filtered_old_value = data_out;
-  old_value = data_in;
+  if (std::isfinite(data_in)) old_value = data_in;
 
   return true;
 }
--- a/control_toolbox/include/control_toolbox/low_pass_filter.hpp
+++ b/control_toolbox/include/control_toolbox/low_pass_filter.hpp
@@ -18,6 +18,7 @@
 #include <Eigen/Dense>
 
 #include <cmath>
+#include <limits>
 #include <memory>
 #include <stdexcept>
 #include <string>
--- a/control_toolbox/include/control_toolbox/low_pass_filter.hpp
+++ b/control_toolbox/include/control_toolbox/low_pass_filter.hpp
@@ -245,7 +245,7 @@ bool LowPassFilter<T>::update(const T & data_in, T & data_out)
   }
   // If this is the first call to update initialize the filter at the current state
   // so that we dont apply an impulse to the data.
-  if (filtered_value == std::numeric_limits<T>::quiet_NaN())
+  if (std::isnan(filtered_value))
   {
     filtered_value = data_in;
     filtered_old_value = data_in;
--- a/control_toolbox/include/control_toolbox/low_pass_filter.hpp
+++ b/control_toolbox/include/control_toolbox/low_pass_filter.hpp
@@ -122,10 +122,8 @@ private:
   // Filter parameters
   double a1_; /** feedbackward coefficient. */
   double b1_; /** feedforward coefficient. */
-  /** internal data storage (double). */
-  double filtered_value, filtered_old_value, old_value;
-  /** internal data storage (std::vector<double>). */
-  std::vector<double> vector_filtered_value, vector_filtered_old_value, vector_old_value;
+  /** internal data storage of template type. */
+  T filtered_value, filtered_old_value, old_value;
   /** internal data storage (wrench). */
   Eigen::Matrix<double, 6, 1> msg_filtered, msg_filtered_old, msg_old;
   bool configured_ = false;
@@ -145,11 +143,11 @@ template <typename T>
 bool LowPassFilter<T>::configure()
 {
   // Initialize storage Vectors
-  filtered_value = filtered_old_value = old_value = 0;
+  filtered_value = filtered_old_value = old_value = std::numeric_limits<T>::quiet_NaN();
   // TODO(destogl): make the size parameterizable and more intelligent is using complex types
   for (Eigen::Index i = 0; i < 6; ++i)
   {
-    msg_filtered[i] = msg_filtered_old[i] = msg_old[i] = 0;
+    msg_filtered[i] = msg_filtered_old[i] = msg_old[i] = std::numeric_limits<double>::quiet_NaN();
   }
 
   return configured_ = true;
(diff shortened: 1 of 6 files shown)


control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -186,6 +188,35 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
...
+template <>
+inline bool LowPassFilter<std::vector<double>>::update(
+  const std::vector<double> & data_in, std::vector<double> & data_out)
+{
+  if (!configured_)
+  {
+    throw std::runtime_error("Filter is not configured");
+  }
+  // If this is the first call to update initialize the filter at the current state
+  // so that we dont apply an impulse to the data.
```
MarqRazz, line 200: Shouldn't the other filter's not initialize to zero in [configure()](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L145-L151)? If I didn't do this I get an impulse applied to my feedback on startup which ends up causing my robot to move.

christophfroehlich: Sounds reasonable. Do you want to fix that, maybe by configuring it with NaN and then set the appropriate initial value in the first update() call?

MarqRazz:
> Implementation-wise LGTM. Do you mind adding the load test [here](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/test/control_filters/test_load_low_pass_filter.cpp), and a simple test [here](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/test/control_filters/test_low_pass_filter.cpp#L88)?

Sure thing. Just wanted to get the initial changes out there to see what you guys thought.

christophfroehlich:
In principle we have this missing in every filter, which is applicable to a double. Maybe we should use the upstream MultiChannelFilter instead? https://github.com/ros/filters/blob/ros2/include/filters/filter_base.hpp#L285C7-L285C29

MarqRazz:
> In principle we have this missing in every filter, which is applicable to a double. Maybe we should use the upstream MultiChannelFilter instead? https://github.com/ros/filters/blob/ros2/include/filters/filter_base.hpp#L285C7-L285C29

If we did that we would need a new header and class for each filter type correct? For example:
```
template <typename T>
class LowPassMultiChannelFilter : public filters::MultiChannelFilterBase<T>
```

What does MultiChannel buy us over the current approach? You still need to write `bool update(const std::vector<T> & data_in, std::vector<T> & data_out)` for each type.

Technically in the hardware_interface I'm writing I am directly using the [control_toolbox::LowPassFilter](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L76) because I don't have any access to the node the interface is running in and that seems to be a requirement to use a `control_filters::LowPassFilter`.

That is I assume because my hw_interface doesn't have access to [this ROS stuff](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/include/control_filters/low_pass_filter.hpp#L94-L96) it segfaults if I try to call [configure()](https://github.com/ros-controls/control_toolbox/blob/ros2-master/control_toolbox/include/control_filters/low_pass_filter.hpp#L101) on a `control_filters::LowPassFilter` object.
```
[taskset-1] #0    Object "/root/c3p0_ws/build/kortex3_driver/libkortex3_driver.so", at 0x72bcfba43d25, in control_filters::LowPassFilter<std::vector<double, std::allocator<double> > >::configure()
[taskset-1] Segmentation fault (Address not mapped to object [(nil)])

File: control_toolbox/include/control_toolbox/low_pass_filter.hpp [hunk 2]
```diff
@@ -124,6 +124,8 @@ class LowPassFilter
   double b1_; /** feedforward coefficient. */
   /** internal data storage (double). */
   double filtered_value, filtered_old_value, old_value;
+  /** internal data storage (std::vector<double>). */
+  std::vector<double> vector_filtered_value, vector_filtered_old_value, vector_old_value;
```
christophfroehlich, line 128: can't we use the template type here, instead of adding additional member variables?

saikishor: I agree. Much better

saikishor: We should alsp start thinking of adding underscore to the member variables

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -186,6 +188,35 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
...
+  }
+  // If this is the first call to update initialize the filter at the current state
+  // so that we dont apply an impulse to the data.
+  // This also sets the size of the member variables to match the input data.
+  if (vector_filtered_value.empty())
+  {
+    vector_filtered_value = data_in;
+    vector_filtered_old_value = data_in;
+    vector_old_value = data_in;
+  }
```
saikishor, line 207:
```suggestion
  }
  else
  {
    assert(data_in.size() == vector_filtered_value.size(), "Internal data and the data_in doesn't hold the same size");
    assert(data_out.size() == data_in.size(), "data_in and data_out doesn't hold same size");
  }
```

saikishor (commented):
Apart from the minor comments, @MarqRazz have you considered using `Eigen::VectorXd` instead of `std::vector`? I think it is better to use `Eigen::VectorXd`

Raivias (approved):
LGTM other than the changes that have already been requested.

May want to consider wrapping one update with the other so there's only one source of logic. It's unusual but I'd suggest wrapping the vector version in the original. i.e. put the value in a vector then feed it to the vector based updater. It will create a single source of logic. You also get the benefit of doing checks once, instead of every index like you would if you wrapped the other way.

MarqRazz:
> Apart from the minor comments, @MarqRazz have you considered using `Eigen::VectorXd` instead of `std::vector`? I think it is better to use `Eigen::VectorXd`

So internally you think it would be better to convert between Eigen and std::vector on each update()? What does switching to Eigen add for us here?

MarqRazz:
The `control_filters` tests that include ROS seem to be flaky. I can get `100% tests passed, 0 tests failed out of 12` but frequently the PID and G-comp test timeout. I think it might be because multiple tests are running at the same time and conflicting with each other.

```
 5/12 Test  #5: test_gravity_compensation ........***Timeout  60.02 sec
test 6
      Start  6: test_load_gravity_compensation

6: Test command: /usr/bin/python3 "-u" "/opt/ros/jazzy/share/ament_cmake_test/cmake/run_test.py" "/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml" "--package-name" "control_toolbox" "--output-file" "/root/colcon_ws/build/control_toolbox/ament_cmake_gmock/test_load_gravity_compensation.txt" "--command" "/root/colcon_ws/build/control_toolbox/test_load_gravity_compensation" "--gtest_output=xml:/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml"
6: Working Directory: /root/colcon_ws/build/control_toolbox
6: Test timeout computed to be: 60
6: -- run_test.py: invoking following command in '/root/colcon_ws/build/control_toolbox':
6:  - /root/colcon_ws/build/control_toolbox/test_load_gravity_compensation --gtest_output=xml:/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml
6: Running main() from gmock_main.cc

The following tests FAILED:
	  5 - test_gravity_compensation (Timeout)
```

christophfroehlich:
> The `control_filters` tests that include ROS seem to be flaky. I can get `100% tests passed, 0 tests failed out of 12` but frequently the PID and G-comp test timeout. I think it might be because multiple tests are running at the same time and conflicting with each other.
> 
> ```
>  5/12 Test  #5: test_gravity_compensation ........***Timeout  60.02 sec
> test 6
>       Start  6: test_load_gravity_compensation
> 
> 6: Test command: /usr/bin/python3 "-u" "/opt/ros/jazzy/share/ament_cmake_test/cmake/run_test.py" "/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml" "--package-name" "control_toolbox" "--output-file" "/root/colcon_ws/build/control_toolbox/ament_cmake_gmock/test_load_gravity_compensation.txt" "--command" "/root/colcon_ws/build/control_toolbox/test_load_gravity_compensation" "--gtest_output=xml:/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml"
> 6: Working Directory: /root/colcon_ws/build/control_toolbox
> 6: Test timeout computed to be: 60
> 6: -- run_test.py: invoking following command in '/root/colcon_ws/build/control_toolbox':
> 6:  - /root/colcon_ws/build/control_toolbox/test_load_gravity_compensation --gtest_output=xml:/root/colcon_ws/build/control_toolbox/test_results/control_toolbox/test_load_gravity_compensation.gtest.xml
> 6: Running main() from gmock_main.cc
> 
> The following tests FAILED:
> 	  5 - test_gravity_compensation (Timeout)
> ```
Try with a sequential executor
https://colcon.readthedocs.io/en/released/reference/executor-arguments.html
https://github.com/ros-controls/ros2_control_ci/blob/master/.github/workflows/reusable-ros-tooling-source-build.yml#L93-L98

MarqRazz:
> Try with a sequential executor

I tried and it didn't help so I could be wrong about what is hanging, but it is flaky. Can you enable the tests and lets see how it goes?
`colcon test --executor sequential --event-handlers console_direct+ --packages-select control_toolbox --pytest-args -s`

I was surprised that I needed to declare my [parameters](https://github.com/MarqRazz/control_toolbox/blob/low_pass_vector/control_toolbox/test/control_filters/test_low_pass_filter.cpp#L165-L167) in testing the vector double calculation and it isn't required in the [others](https://github.com/MarqRazz/control_toolbox/blob/low_pass_vector/control_toolbox/test/control_filters/test_low_pass_filter.cpp#L35).

christophfroehlich (commented):
minor suggestion, otherwise this looks fine!

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -186,13 +200,57 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
...
 template <typename T>
 bool LowPassFilter<T>::update(const T & data_in, T & data_out)
 {
   if (!configured_)
   {
     throw std::runtime_error("Filter is not configured");
   }
+  // If this is the first call to update initialize the filter at the current state
+  // so that we dont apply an impulse to the data.
+  if (filtered_value == std::numeric_limits<T>::quiet_NaN())
```
christophfroehlich, line 248:
```suggestion
  if (std::isnan(filtered_value))
```

MarqRazz:
Thanks for the quick responses guys! :heart: 

Could we please back port to Jazzy :pray:

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -186,13 +200,57 @@ inline bool LowPassFilter<geometry_msgs::msg::WrenchStamped>::update(
...
+      "Internal data and the data_in doesn't hold the same size");
+    assert(data_out.size() == data_in.size() && "data_in and data_out doesn't hold same size");
+  }
+
+  // Filter each value in the vector
+  for (std::size_t i = 0; i < data_in.size(); i++)
+  {
+    data_out[i] = b1_ * old_value[i] + a1_ * filtered_old_value[i];
+    filtered_old_value[i] = data_out[i];
+    old_value[i] = data_in[i];
```
saikishor, line 233:
```suggestion
    if(!std::finite(data_in[i]))
    {
      old_value[i] = data_in[i];
    }
```
What do you think about protecting the filter from any NaNs?

saikishor (commented):
Looks good to me, just a minor thing

christophfroehlich (approved):

saikishor (approved):
LGTM

PR #347, opened 2025-04-24
Minor filter fixes + clang-format rules update

saikishor:
A follow-up of https://github.com/ros-controls/control_toolbox/pull/344


.clang-format
christophfroehlich, line 1: do we have to update in all the repos?

saikishor: Not sure 😅😅

christophfroehlich (approved):
Thanks!

Issue #345, opened 2025-04-24
Improve member variables of low_pass_filter


christophfroehlich:
## Background

Currently, there are two member variables for different templated specialization (`wrench`, and all others use `T`)
https://github.com/ros-controls/control_toolbox/blob/b1f2619ef54a1ba3fdb38cb9dcce1745dd6ac31b/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L126-L129

Improve that by only adding a single variant depending on the template argument.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html), for Step 3 use "Download Source Code" section with [these instructions](https://control.ros.org/master/doc/getting_started/getting_started.html#building-from-source).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

- [ ] 💾 **Commit and Push** your changes

- [ ] 🔀 **Start a Pull Request** to request to merge your code into `master`. There are two ways that you can start a pull request:
[...]

PR #351, opened 2025-04-27
Make member variables of LowPassFilter class generic

pedroazeredo04:
This PR aims to tackle #345.

In order to do so,  I followed the advice from @christophfroehlich, using [`type_traits`](https://en.cppreference.com/w/cpp/header/type_traits) to identify which type was passed to the template of the `LowPassFilter` and configure the `StorageType` accordingly, all of that in compile time.

Since there is now only one set of three variables instead of two sets, I also corrected all the other occurrences of the variables were removed.

Also, @christophfroehlich said

> as an alternative, a traits struct with an initialize() method and operator overloads with a specialization for every non-trivial msg type could work and make the implementation itself cleaner. (no overload of update() would be necessary)

I thought this was a very beautiful idea, but I had some difficulties making this actually happen, because the function `update()` has some particularities for each data type. At the end, I was basically re-implementing the `update()` function three times within the traits struct. So I thought it was simpler and more readable to leave the overload the way it is. However, I am open to suggestions!!

One last thing: I had to change the base branch to `jazzy`, because the `ros2-master` branch was not compiling in my machine. Would you like for me to target the branch `ros2-master` instead of `jazzy`?

Please let me know anything that would be nice to change


saikishor (changes requested):
@pedroazeredo04 can you fix the pre-commit?. Just install pre-commit and run `pre-commit run --all-files`

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -123,10 +124,17 @@ class LowPassFilter
...
   double a1_; /** feedbackward coefficient. */
   double b1_; /** feedforward coefficient. */
-  /** internal data storage of template type. */
-  T filtered_value, filtered_old_value, old_value;
-  /** internal data storage (wrench). */
-  Eigen::Matrix<double, 6, 1> msg_filtered, msg_filtered_old, msg_old;
+
+  // Define the storage type based on T
+  using StorageType = typename std::conditional<
+    std::is_same<T, geometry_msgs::msg::WrenchStamped>::value, Eigen::Matrix<double, 6, 1>,
```
saikishor, line 130:
```suggestion
    std::is_same<T, geometry_msgs::msg::WrenchStamped>::value, Eigen::Vector6d,
```
May be we could use this directly

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -143,12 +151,18 @@ LowPassFilter<T>::~LowPassFilter()
...
-  // TODO(destogl): make the size parameterizable and more intelligent is using complex types
-  for (Eigen::Index i = 0; i < 6; ++i)
+  // Initialize storage Vectors, depending on the type of T
+  if constexpr (std::is_same<T, geometry_msgs::msg::WrenchStamped>::value)
+  {
+    // TODO(destogl): make the size parameterizable and more intelligent is using complex types
+    for (Eigen::Index i = 0; i < 6; ++i)
+    {
+      filtered_value_[i] = filtered_old_value_[i] = old_value_[i] = std::numeric_limits<double>::quiet_NaN();
+    }
```
saikishor, line 161: This can be completely skipped, if we define EIGEN_INITIALIZE_MATRICES_BY_NAN before including Eigen. @christophfroehlich what do you think about it?. I don't have a strong opinion here

saikishor (changes requested):

control_toolbox/include/control_toolbox/low_pass_filter.hpp
christophfroehlich: didn't know about that, why not

control_toolbox/include/control_toolbox/low_pass_filter.hpp
christophfroehlich: This might come from 
https://github.com/ros/geometry/blob/noetic-devel/eigen_conversions/include/eigen_conversions/eigen_msg.h#L100
according to the comment
`  // TODO(destogl): use wrenchMsgToEigen`
But I don't find these conversion wrappers for ROS 2, https://docs.ros.org/en/rolling/p/tf2_eigen/generated/index.html#tf2-eigen

christophfroehlich:
Thanks a lot for your contribution.

> I thought this was a very beautiful idea, but I had some difficulties making this actually happen, because the function `update()` has some particularities for each data type. At the end, I was basically re-implementing the `update()` function three times within the traits struct. So I thought it was simpler and more readable to leave the overload the way it is. However, I am open to suggestions!!

You are right about that, I made a proposal in the issue: what do you think? I still think that the algorithms in the filters would be more clean, and we easily could apply that to all the other filters and add for example Pose messages (in a future PR).

pedroazeredo04:
@christophfroehlich I commited your suggestion of having a `FilterTrait` struct! But I still couldn't quite make the function `LowPassFilter::update()` have an unique implementation for the three distinct data types :( 

That is because the override of this function for the type `std::vector<double>` does some additional verifications that the other data types doesn't. For instance, this block

```C
if (filtered_value_.empty())
  {
    if (std::any_of(data_in.begin(), data_in.end(), [](double val) { return !std::isfinite(val); }))
    {
      return false;
    }
    filtered_value_ = data_in;
    filtered_old_value_ = data_in;
    old_value_ = data_in;
  }
  else
  {
    assert(
      data_in.size() == filtered_value_.size() &&
      "Internal data and the data_in doesn't hold the same size");
    assert(data_out.size() == data_in.size() && "data_in and data_out doesn't hold same size");
  }
```

is very distinct from the other overrides, so it is hard to generalize this for every data type. 

---

Also, about the other methods of the `TypeTraits` struct, I could add the `isnan()`, `isempty()` and `copy_metadata()` if you like, what do you think?

> Something like this, plus other necessary member functions like `isnan`, `isempty`, `copy_metadata` (for copying the header if it is a message type) or operator overloads for `+operator`, `*operator`, which can go in its own header file:

I would only have to say , in order to do an operator overload for `+operator` and `*operator` for every data type, I think we would have to create a wrapper struct for each non-primitive type, because we probably shouldn't overload the `+operator` of `std::vector<double>`, for example, as it is [undefined behavior](https://en.cppreference.com/w/cpp/language/extending_std). What I mean is that instead of 

```C
  // undefined behavior
  template<typename T>
  std::vector<T> operator+(const std::vector<T>& lhs, const std::vector<T>& rhs) {
    std::vector<T> result(lhs.size());
    for (size_t i = 0; i < lhs.size(); ++i) {
      result[i] = lhs[i] + rhs[i];
    }
    return result;
  }
```

We would probably have to do

```C
// Wrapper around std::vector<double> to enable element-wise + and * without touching std::
struct VecD {
  std::vector<double> v;
  VecD() = default;
  VecD(const std::vector<double>& data) : v(data) {}
  bool empty() const { return v.empty(); }
  void resize(size_t n) { v.resize(n); }
  double& operator[](size_t i) { return v[i]; }
  const double& operator[](size_t i) const { return v[i]; }
};

// element-wise add
inline VecD operator+(const VecD& a, const VecD& b) {
  VecD out(a.v.size());
  for (size_t i = 0; i < a.v.size(); ++i)
    out.v[i] = a.v[i] + b.v[i];
  return out;
}

```

christophfroehlich:
Thanks for your work on this. My suggestion was only a draft, I'm not really experienced in writing templated classes, so I might be running in a dead end. Please shout out if this is the case ;)

> @christophfroehlich I commited your suggestion of having a `FilterTrait` struct! But I still couldn't quite make the function `LowPassFilter::update()` have an unique implementation for the three distinct data types :(
> 
> That is because the override of this function for the type `std::vector<double>` does some additional verifications that the other data types doesn't. For instance, this block
> 
> ```c
> if (filtered_value_.empty())
>   {
>     if (std::any_of(data_in.begin(), data_in.end(), [](double val) { return !std::isfinite(val); }))
>     {
>       return false;
>     }
>     filtered_value_ = data_in;
>     filtered_old_value_ = data_in;
>     old_value_ = data_in;
>   }
>   else
>   {
>     assert(
>       data_in.size() == filtered_value_.size() &&
>       "Internal data and the data_in doesn't hold the same size");
>     assert(data_out.size() == data_in.size() && "data_in and data_out doesn't hold same size");
>   }
> ```
> 
> is very distinct from the other overrides, so it is hard to generalize this for every data type.

The if branch could be implemented with the already mentionend isempty(), isnan(), and the else branch could be solved by a `if constexpr` on std:.vector?
```c++
// helper: default false
template<typename T>
struct is_std_vector : std::false_type {};

// specialization for std::vector
template<typename U, typename Alloc>
struct is_std_vector<std::vector<U, Alloc>> : std::true_type {};
```
and use     `if constexpr (is_std_vector<T>::value) {}`

> Also, about the other methods of the `TypeTraits` struct, I could add the `isnan()`, `isempty()` and `copy_metadata()` if you like, what do you think?
> 
> > Something like this, plus other necessary member functions like `isnan`, `isempty`, `copy_metadata` (for copying the header if it is a message type) or operator overloads for `+operator`, `*operator`, which can go in its own header file:
> 
> I would only have to say , in order to do an operator overload for `+operator` and `*operator` for every data type, I think we would have to create a wrapper struct for each non-primitive type, because we probably shouldn't overload the `+operator` of `std::vector<double>`, for example, as it is [undefined behavior](https://en.cppreference.com/w/cpp/language/extending_std). What I mean is that instead of

This is a very good hint (haven't thought about that), but we aren't overloading this to `std::vector<double>`, but to struct `FilterTraits<std::vector<T>>`? Isn't this a difference w.r.t to your concern?

pedroazeredo04:
Hi @christophfroehlich ! No need to thank my work on this, I'm having fun messing around with templates :)

I pushed a whip code that tries to make the update() function trully generic.

> The if branch could be implemented with the already mentionend isempty(), isnan(), and the else branch could be solved by a `if constexpr` on std:.vector?
> 
> ```c++
> // helper: default false
> template<typename T>
> struct is_std_vector : std::false_type {};
> 
> // specialization for std::vector
> template<typename U, typename Alloc>
> struct is_std_vector<std::vector<U, Alloc>> : std::true_type {};
> ```
> 
> and use `if constexpr (is_std_vector<T>::value) {}`

That is a good idea! I did something similar, but checking for the types inside of the function itself.

 > This is a very good hint (haven't thought about that), but we aren't overloading this to std::vector<double>, but to struct FilterTraits<std::vector<T>>? Isn't this a difference w.r.t to your concern?

In my understanding, the problem is that `FilterTraits<std::vector<T>>` isn't the type of the variables `filtered_value_`, `old_value_` and `filtered_old_value_`. Those variables _are_ `std::vector<double>`. So, in order for the block

```C
  filtered_value_ = old_value_ * b1_ + filtered_old_value_ * a1_;
  filtered_old_value_ = filtered_value_;
```

work for every possible datatype, I had to add in a `FilterVector` wrapper around the `std::vector`. So, for the `std::vector<double>` overload, the `StorageType` is actually of type `FilterVector`, and not `std::vector<double>`. At the end I though it turned out not so ugly ahahaha, but I'm of course open to suggestions. If you could review it and give your feedback, I would appreciate it!

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,198 @@
...
+  using StorageType = T;
+
+  static void initialize(StorageType & storage)
+  {
+    storage = T{std::numeric_limits<T>::quiet_NaN()};
+  }
+
+  static bool is_nan(const StorageType & storage) { return std::isnan(storage); }
+
+  static bool is_infinite(const StorageType & storage) { return !std::isfinite(storage); }
```
christophfroehlich, line 87: i'd prefer to use the `is_finite` instead of `is_infinite`, as we are used to this from std

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,198 @@
...
+    return false;
+  }
+
+  static void assign(StorageType & storage, const StorageType & data_in) { storage = data_in; }
+};
+
+template <>
+struct FilterTraits<geometry_msgs::msg::WrenchStamped>
+{
+  using StorageType = Eigen::Matrix<double, 6, 1>;
```
christophfroehlich, line 101: should we use?
```suggestion
  using StorageType = Vector6d;
```

christophfroehlich (changes requested):
I like the direction of the simplification of the filter itself. The filter_traits look a bit verbose, but the API is clean and understandable IMHO.

Only some minor comments from my side left

control_toolbox/include/control_toolbox/low_pass_filter.hpp
```diff
@@ -143,128 +149,59 @@ LowPassFilter<T>::~LowPassFilter()
...
-    assert(data_out.size() == data_in.size() && "data_in and data_out doesn't hold same size");
-  }
-
-  // Filter each value in the vector
-  for (std::size_t i = 0; i < data_in.size(); i++)
-  {
-    data_out[i] = b1_ * old_value[i] + a1_ * filtered_old_value[i];
-    filtered_old_value[i] = data_out[i];
-    if (std::isfinite(data_in[i]))
+    if constexpr (std::is_same_v<T, std::vector<double>>)
```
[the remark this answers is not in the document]
pedroazeredo04: Do you think I should change that to generalize the type of the vector? If I understood correctly, I believe the previous code only supported `std::vector<double` also

control_toolbox/include/control_toolbox/filter_traits.hpp
pedroazeredo04: I didn't found the `Eigen::Vector6d` type, and the compiler doesn't seem to like when I change to this type. Should I include some specific file in order to use this type?

pedroazeredo04:
Thanks for the review @christophfroehlich !!

I commited some the small fixes that you suggested, and I will try to fix the failing tests later.

control_toolbox/include/control_toolbox/low_pass_filter.hpp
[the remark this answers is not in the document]
christophfroehlich: have you tried my code snippet above? I thought this could work. 
we could also make another method for "validate_input" or similar? could be useful for other message types also, like checking if the frame_id the same or whatever.

pedroazeredo04:
Hi @christophfroehlich. I had some time today and I added the generic method `Traits::validate_input` to the `FilterTraits` struct, as you suggested. I also found a bug in the `Eigen::Matrix` initialization, and fixed that too. Hopefully, with those changes, all tests will pass😄. If you could take a look at the code and give me your feedback when you have a chance, that would be nice!

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+// helper: default false
+template <typename T>
+struct is_std_vector : std::false_type
+{
+};
+
+// specialization for std::vector
+template <typename U, typename Alloc>
+struct is_std_vector<std::vector<U, Alloc>> : std::true_type
```
christophfroehlich, line 42: is this used now?

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  static void add_metadata(DataType & data_out, const DataType & data_in)
+  {
+    data_out.header = data_in.header;
+  }
+};
+
+template <>
+struct FilterTraits<std::vector<double>>
```
christophfroehlich, line 190: can this be generalized for all types inside a std::vector?

christophfroehlich (approved):
LGTM, 2 minor comments

christophfroehlich (approved):

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,246 @@
...
+    (void)data_in;
+    (void)filtered_value;
+    (void)data_out;  // Suppress unused warnings
+  }
+
+  static void add_metadata(StorageType & storage, const StorageType & data_in)
+  {
+    (void)storage;
+    (void)data_in;
+  }
```
saikishor, line 113: Can't we make it pure virtual?

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+  static void add_metadata(DataType & data_out, const DataType & data_in)
+  {
+    data_out.header = data_in.header;
+  }
+};
+
+template <typename U, typename Alloc>
+struct FilterTraits<std::vector<U, Alloc>>
```
saikishor, line 183: Are we specialising with allocator here?

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,246 @@
...
+    FilterVector result = *this;
+    for (size_t i = 0; i < data.size(); ++i)
+    {
+      result.data[i] += other.data[i];
+    }
+    return result;
+  }
+
+  size_t size() const { return data.size(); }
+};
```
saikishor, line 71: Are we using this somewhere?

control_toolbox/include/control_toolbox/filter_traits.hpp
pedroazeredo04: If I understand it correctly this function will only be used by "default" vectors, so I believe we could just drop the `Alloc`, what do you think?

control_toolbox/include/control_toolbox/filter_traits.hpp
pedroazeredo04: This is being used as the [`StorageType`](https://github.com/ros-controls/control_toolbox/blob/56c7edbff44b413cb449cb9ef4ded660bcf28429/control_toolbox/include/control_toolbox/filter_traits.hpp#L185) of the `std::vector` overload of the `FilterTraits`.

The reason for that is that we want to override `operator+` and `operator*` for every DataType, so that the implementation inside [`low_pass_filter.hpp`](https://github.com/ros-controls/control_toolbox/blob/56c7edbff44b413cb449cb9ef4ded660bcf28429/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L184-L185) stays the same:

```C
  // Filter
  filtered_value_ = old_value_ * b1_ + filtered_old_value_ * a1_;
  filtered_old_value_ = filtered_value_;
```

However, we cannot override `operator+` and `operator*` of `std::vector`, because that leads to undefined behavior. So I wrote a wrapper around the std::vector just so that we could override these operators.

An alternative for that would be to write an `add()` and `multiply()` method for every overload of the `FilterTraits`. But I think that would be not so nice, as the above snippet for the filter would not be so legible, because we would have to do something like: 

```C
filtered_value_ = add(multiply(old_value_, b1_), multiply(filtered_old_value_, a1_));
```

So that is why that wrapper is there.

control_toolbox/include/control_toolbox/filter_traits.hpp
pedroazeredo04: I think we can't do virtual templates (I say this based on [this discussion](https://stackoverflow.com/questions/2354210/can-a-class-member-function-template-be-virtual) here). But the discussion does mention type erasure as an alternative, which might be applicable in our case. What do you think?

control_toolbox/include/control_toolbox/filter_traits.hpp
saikishor: Yes! You are right. That's why I asked about the alloc template argument

christophfroehlich:
This happens if you push a new commit after a review, no worries

control_toolbox/include/control_toolbox/filter_traits.hpp
saikishor: Got it

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+template <typename U>
+struct FilterTraits<std::vector<U>>
+{
+  using StorageType = FilterVector<U>;
+  using DataType = std::vector<U>;
+
+  static void initialize(StorageType & storage)
+  {
+    storage.data = std::vector<U>{std::numeric_limits<U>::quiet_NaN()};
```
saikishor, line 190: I'm thinking about this one. I'm not sure if we have to initialize it or not, because in the first loop it will have to fill in the data right?. Do we want to have it pre-filled with the data?

I'm just saying because we are not sure about the size at the initialization time

saikishor (approved):
LGTM. Just a minor comment

control_toolbox/include/control_toolbox/filter_traits.hpp
pedroazeredo04: That's true, I think it would be better to leave it empty.

In the **[`low_pass_filter.hpp`](https://github.com/pedroazeredo04/control_toolbox/blob/f68ad4a3f489a71646a42fa44023347783e8e8da/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L166),** we already check for emptiness to check if it is a first call. So it would make sense to leave the `std::vector` empty, because, as you said, we don't know its size at the initialization time.

But after your question I thought: does it even make sense to have a [`configure()`](https://github.com/pedroazeredo04/control_toolbox/blob/f68ad4a3f489a71646a42fa44023347783e8e8da/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L147-L155) method to make the filter variables be NaN? I think it would be simpler if we removed the `configure()` method for all overloads, not only the `std::vector`.

And we could check if it is the first time that the `update()` method is being called by a member `bool first_run_` or something like that. So instead of having:

```C
bool LowPassFilter<T>::update(const T & data_in, T & data_out)
{
  if (!configured_)
  {
    throw std::runtime_error("Filter is not configured");
  }
  // If this is the first call to update initialize the filter at the current state
  // so that we dont apply an impulse to the data.
  if (Traits::is_nan(filtered_value_) || Traits::is_empty(filtered_value_))
  {
  // rest of the code
  }
}
```

We could have:

```C
bool LowPassFilter<T>::update(const T & data_in, T & data_out)
{
  if (!configured_)
  {
    throw std::runtime_error("Filter is not configured");
  }
  // If this is the first call to update initialize the filter at the current state
  // so that we dont apply an impulse to the data.
  if (first_run_)
  {
   // rest of the code
  first_run_ = false
  }
}

```

And remove the `configure()` method entirely. What do you think?

saikishor:
```suggestion
    (void)storage;
```

pedroazeredo04: It is done!

saikishor: LGTM. let me run the CI

saikishor (approved):

christophfroehlich (approved):
Thanks again, LGTM!

PR #387, opened 2025-06-06
Use RealtimeThreadSafeBox for PID class

christophfroehlich:
https://github.com/ros-controls/realtime_tools/pull/347

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -788,9 +788,6 @@ public:
   }
 
 protected:
-  // Store the PID gains in a realtime box to allow dynamic reconfigure to update it without
-  // blocking the realtime update loop
-  realtime_tools::RealtimeThreadSafeBox<Gains> gains_box_;
   // local copy of the gains for the RT loop
   Gains gains_{
     0.0,
@@ -799,6 +796,9 @@ protected:
     std::numeric_limits<double>::infinity(),
     -std::numeric_limits<double>::infinity(),
     AntiWindupStrategy()};
+  // Store the PID gains in a realtime box to allow dynamic reconfigure to update it without
+  // blocking the realtime update loop
+  realtime_tools::RealtimeThreadSafeBox<Gains> gains_box_{gains_};
 
   double p_error_last_ = 0; /** Save state for derivative state calculation. */
   double p_error_ = 0;      /** Error. */


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -560,9 +566,19 @@ class Pid
...
+   *
+   * \note This method is not RT safe
    */
   Gains get_gains();
 
+  /*!
+   * \brief Get PID gains for the controller.
+   * \return gains A struct of the PID gain values
+   *
+   * \note This method must be called from the RT loop only
```
saikishor:
```suggestion
   * \note This method can be called from the RT loop
```

control_toolbox/src/pid.cpp
```diff
@@ -158,6 +157,9 @@ void Pid::reset(bool save_i_term)
   {
     clear_saved_iterm();
   }
+
+  // blocking, as reset() is not called in the RT thread
+  gains_ = gains_box_.get();
```
saikishor, line 143: Do we need this in the `reset` method?. Is that to have a valid gains in the compute_command on startup?

The only thing is when this is written in the nonRT loop (post startup) we might be writing at the same time when the RT loop is reading. This might not be ideal

saikishor (commented):
The changes are fine but I've some questions. I've mentioned those comments. 

BTW why not use the RTThreadSafeBox Directly?. As we are using the realtime mutexes, it is safe. This way we can avoid another variable completely

control_toolbox/src/pid.cpp
christophfroehlich: gain_ variable is never initialized otherwise? if the try_get in compute_command fails, then the gains_ are not valid.

christophfroehlich:
> BTW why not use the RTThreadSafeBox Directly?. As we are using the realtime mutexes, it is safe. This way we can avoid another variable completely

what do you mean? instead of the gain_ copy? but what happens if try_get in the compute_command fails, then we can't proceed?

saikishor (approved):

control_toolbox/src/pid.cpp
```diff
@@ -50,7 +50,7 @@ constexpr double UMAX_INFINITY = std::numeric_limits<double>::infinity();
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 Pid::Pid(double p, double i, double d, double i_max, double i_min, bool antiwindup)
-: gains_buffer_()
+: gains_(0.0, 0.0, 0.0, 0.0, 0.0, AntiWindupStrategy())
```
saikishor:
```suggestion
: gains_(0.0, 0.0, 0.0, std::numeric_limits<double>::infinity(), -std::numeric_limits<double>::infinity(), AntiWindupStrategy())
```

control_toolbox/src/pid.cpp
```diff
@@ -73,7 +73,7 @@ Pid::Pid(double p, double i, double d, double i_max, double i_min, bool antiwind
 Pid::Pid(
   double p, double i, double d, double u_max, double u_min,
   const AntiWindupStrategy & antiwindup_strat)
-: gains_buffer_()
+: gains_(0.0, 0.0, 0.0, 0.0, 0.0, AntiWindupStrategy())
```
saikishor:
```suggestion
: gains_(0.0, 0.0, 0.0, std::numeric_limits<double>::infinity(), -std::numeric_limits<double>::infinity(), AntiWindupStrategy())
```

control_toolbox/src/pid.cpp
```diff
@@ -87,10 +87,10 @@ Pid::Pid(
   reset();
 }
 
-Pid::Pid(const Pid & source)
+Pid::Pid(const Pid & source) : gains_(0.0, 0.0, 0.0, 0.0, 0.0, AntiWindupStrategy())
```
saikishor:
```suggestion
Pid::Pid(const Pid & source) : gains_(0.0, 0.0, 0.0, std::numeric_limits<double>::infinity(), -std::numeric_limits<double>::infinity(), AntiWindupStrategy())
```

saikishor (approved):

saikishor (approved):

PR #390, opened 2025-06-08
Cleanup saturation parameter in the methods and constructors

saikishor:
This PR consists of changes to cleanup the saturation boolean of the PID toolbox

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -276,6 +276,8 @@ public:
           //   std::to_string(u_max) + ")");
           std::cout << "Received invalid u_min and u_max values: " << "u_min: " << u_min
                     << ", u_max: " << u_max << ". Setting saturation to false." << std::endl;
+          u_max_ = std::numeric_limits<double>::infinity();
+          u_min_ = -std::numeric_limits<double>::infinity();
           saturation_ = false;
         }
       }
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -258,28 +258,22 @@ public:
       u_max_(u_max),
       u_min_(u_min),
       trk_tc_(trk_tc),
-      saturation_(false),
       antiwindup_(antiwindup),
       antiwindup_strat_(antiwindup_strat)
     {
-      saturation_ = std::isfinite(u_min) || std::isfinite(u_max);
-      if (saturation_)
+      if (std::isnan(u_min) || std::isnan(u_max))
       {
-        if (std::isnan(u_min) || std::isnan(u_max))
-        {
-          throw std::invalid_argument("Gains: u_min and u_max must not be NaN");
-        }
-        if (u_min > u_max)
-        {
-          // throw std::invalid_argument(
-          //   "Gains: u_min (" + std::to_string(u_min) + ") must not be greater than u_max (" +
-          //   std::to_string(u_max) + ")");
-          std::cout << "Received invalid u_min and u_max values: " << "u_min: " << u_min
-                    << ", u_max: " << u_max << ". Setting saturation to false." << std::endl;
-          u_max_ = std::numeric_limits<double>::infinity();
-          u_min_ = -std::numeric_limits<double>::infinity();
-          saturation_ = false;
-        }
+        throw std::invalid_argument("Gains: u_min and u_max must not be NaN");
+      }
+      if (u_min > u_max)
+      {
+        // throw std::invalid_argument(
+        //   "Gains: u_min (" + std::to_string(u_min) + ") must not be greater than u_max (" +
+        //   std::to_string(u_max) + ")");
+        std::cout << "Received invalid u_min and u_max values: " << "u_min: " << u_min
+                  << ", u_max: " << u_max << ". Setting saturation to false." << std::endl;
+        u_max_ = std::numeric_limits<double>::infinity();
+        u_min_ = -std::numeric_limits<double>::infinity();
       }
     }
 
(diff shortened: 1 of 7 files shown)


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,10 +257,21 @@ class Pid
       u_max_(u_max),
       u_min_(u_min),
       trk_tc_(trk_tc),
-      saturation_(saturation),
+      saturation_(false),
       antiwindup_(antiwindup),
       antiwindup_strat_(antiwindup_strat)
     {
+      if (std::isfinite(u_min) && std::isfinite(u_max) && u_min != 0.0 && u_max != 0.0)
```
christophfroehlich: I'd use an OR instead, it could be a one-sided limit. And why is zero an invalid limit?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,10 +257,21 @@ class Pid
...
+      {
+        saturation_ = true;
```
christophfroehlich: can't we get rid of this variable at all?
https://github.com/ros-controls/control_toolbox/blob/f652c3db485bde684915396e433f588291c985b6/control_toolbox/src/pid.cpp#L407
with +/- infinity, this line would just do nothing?

christophfroehlich (changes requested):

control_toolbox/include/control_toolbox/pid.hpp
saikishor: zero isn't a valid limits I want to cleanup later and wanted to having a working version comitted

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Don't you like to have it?, what I have in mind is like, if someone wants to disable it via dynamic parameters, the user can simply set it to `false`, rather than having to set infinity in the GUI, I found it relatively better in that case

christophfroehlich: I would remove it, my opinion is that 2 parameters is better than 3. And I don't see the benefit of a single param for deactivation. For debugging it does not matter that much, and programmatically it even does not matter how many parameters you have to set?

saikishor: How can we disable the saturation at runtime without such a parameter? I'm thinking about the debugging purposes. I mean anyway the third parameter is set based on the other two implicitly, so shouldn't be a problem right?. 

If you still think the same, sure I can remove it :)

christophfroehlich: what do you mean at runtime? can't you dynamically reconfigure the u_max/u_min parameters?

saikishor: Yes, we can change with rqt_reconfigure but we cannot set the limits to Infinity from the GUI, which means we will never be able to disable them completely. Do I explain myself?

christophfroehlich: Set it to an arbitrary high number, this would have the same effect. But if you insist, I'm also fine with keeping it ;) But then I'd suggest to configure it as "opt-out", which means that saturation=true and u_min=-inf, u_max=inf should be default values. Here, and also in https://github.com/ros-controls/ros2_controllers/pull/1585

control_toolbox/src/pid.cpp
```diff
@@ -406,9 +401,16 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
     cmd_unsat_ = p_term + i_term_ + d_term;
   }
 
-  if (gains.saturation_)
+  if (std::isfinite(gains.u_min_) || std::isfinite(gains.u_max_))
```
christophfroehlich, line 408: do we need this check here? shouldn't that be done in the set_gains method or parameter callback?

saikishor: I just added it here and if somehow it passes the traditional set_gains

christophfroehlich: Shouldn't we move/copy it to: https://github.com/ros-controls/control_toolbox/blob/22ab396f8edf8c65fdb4f951473c988fbc02227e/control_toolbox/src/pid.cpp#L264

but if you think it is still necessary, you can also check it here again

saikishor: ok

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,10 +258,23 @@ class Pid
...
     {
+      if (std::isnan(u_min) || std::isnan(u_max))
+      {
+        throw std::invalid_argument("Gains: u_min and u_max must not be NaN");
+      }
+      if (u_min > u_max)
+      {
+        // throw std::invalid_argument(
+        //   "Gains: u_min (" + std::to_string(u_min) + ") must not be greater than u_max (" +
+        //   std::to_string(u_max) + ")");
```
```suggestion
```

control_toolbox/src/pid.cpp
```diff
@@ -265,11 +265,15 @@ void Pid::set_gains(const Gains & gains_in)
...
   {
-    std::cout << "received i_min > i_max, skip new gains\n";
+    std::cout << "Received i_min > i_max, skip new gains" << std::endl;
   }
   else if (gains_in.u_min_ > gains_in.u_max_)
   {
-    std::cout << "received u_min > u_max, skip new gains\n";
+    std::cout << "Received u_min > u_max, skip new gains" << std::endl;
+  }
+  else if (std::isnan(gains.u_min_) || std::isnan(gains.u_max_))
```
saikishor:
```suggestion
  else if (std::isnan(gains_in.u_min_) || std::isnan(gains_in.u_max_))
```

control_toolbox/src/pid_ros.cpp
```diff
@@ -554,6 +563,39 @@ void PidROS::set_parameter_event_callback()
     /// @note don't use getGains, it's rt
     Pid::Gains gains = pid_.get_gains();
     bool changed = false;
+    // THe saturation parameter is special, it can change the u_min and u_max parameters
```
christophfroehlich:
```suggestion
    // The saturation parameter is special, it can change the u_min and u_max parameters
```

christophfroehlich (approved):
Thanks for the cleanup, LGTM

christophfroehlich (approved):

PR #391, opened 2025-06-09
Don't validate data_out if it is empty

christophfroehlich:
https://github.com/ros-controls/ros2_controllers/pull/560 fails because filter_chains does not initialize intermediate output variables, which makes the assertion `data_in.header.frame_id == data_out.header.frame_id` fail.

I also changed the assertions to throwing exceptions, using fmt to concatenate numbers and strings.

### DIFF
--- a/control_toolbox/include/control_toolbox/filter_traits.hpp
+++ b/control_toolbox/include/control_toolbox/filter_traits.hpp
@@ -17,10 +17,12 @@
 
 #define EIGEN_INITIALIZE_MATRICES_BY_NAN
 
+#include <fmt/core.h>
 #include <Eigen/Dense>
 #include <algorithm>
 #include <cmath>
 #include <limits>
+#include <stdexcept>
 #include <vector>
 
 #include "geometry_msgs/msg/wrench_stamped.hpp"
@@ -58,7 +60,13 @@ struct FilterVector
 
   FilterVector operator+(const FilterVector & other) const
   {
-    assert(data.size() == other.data.size() && "Vectors must be of the same size for addition");
+    if (data.size() != other.data.size())
+    {
+      throw std::runtime_error(
+        fmt::format(
+          "Vectors must be of the same size for addition ({} vs {}).", data.size(),
+          other.data.size()));
+    }
     FilterVector result = *this;
     for (size_t i = 0; i < data.size(); ++i)
     {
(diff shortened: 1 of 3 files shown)


control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -225,10 +237,22 @@ struct FilterTraits<std::vector<U>>
...
+    if (data_in.size() != filtered_value.size())
+    {
+      throw std::runtime_error(
+        fmt::format(
+          "Input vector size ({}) does not match internal state size ({}).", data_in.size(),
+          filtered_value.size()));
+    }
+    // Compare new input's size with output's size,
+    // but only if it existed (filter_chains does not initialize output)
+    if (!data_out.empty() && data_out.size() != data_in.size())
```
saikishor, line 249: Why are we checking if it's empty or not?. I mean the size is important to check right? as we are not resizing it internally

control_toolbox/include/control_toolbox/filter_traits.hpp
```diff
@@ -165,12 +173,16 @@ struct FilterTraits<geometry_msgs::msg::WrenchStamped>
...
-    (void)filtered_value;  // Not used here
+    (void)filtered_value;  // filtered_value has no header
 
-    // Compare new input's frame_id with previous output's frame_id
-    assert(
-      data_in.header.frame_id == data_out.header.frame_id &&
-      "Frame ID changed between filter updates");
+    // Compare new input's frame_id with previous output's frame_id,
+    // but only if it existed (filter_chains does not initialize output)
+    if (!data_out.header.frame_id.empty() && data_in.header.frame_id != data_out.header.frame_id)
```
saikishor, line 180: Can the data_out frame_id be empty?

christophfroehlich: yes, in a filter_chain it is: `buffer0_` will never be initialized. 
https://github.com/ros/filters/blob/e47fe49ef2f3189eb96b3de7714b369e0c51107c/include/filters/filter_chain.hpp#L364

Moreover, in the linked PR in ros2_controllers, we put the wrench message into a low-pass filter, and the frame ID is not important at all.

control_toolbox/include/control_toolbox/filter_traits.hpp
christophfroehlich: same like with frame_id. 
Moreover, in the LPF we only assign the result to the output 
https://github.com/ros-controls/control_toolbox/blob/f652c3db485bde684915396e433f588291c985b6/control_toolbox/include/control_toolbox/low_pass_filter.hpp#L188

saikishor: Got it

control_toolbox/include/control_toolbox/filter_traits.hpp
saikishor: Thanks

saikishor (approved):

PR #400, opened 2025-06-11
Redefine the antiwindup strategy and their parameters interfacing

saikishor:
Second PR for the antiwindup strategy integration

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -74,8 +74,10 @@ public:
     CONDITIONAL_INTEGRATION
   };
 
-  constexpr AntiwindupStrategy() : type(LEGACY) {}
-  constexpr AntiwindupStrategy(Value v) : type(v) {}  // NOLINT(runtime/explicit)
+  AntiwindupStrategy()
+  : type(UNDEFINED), i_min(0.0), i_max(0.0), trk_tc(0.0), legacy_antiwindup(false)
+  {
+  }
 
   void set_type(const std::string & s)
   {
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -258,6 +258,7 @@ public:
       antiwindup_strat_.type = AntiwindupStrategy::LEGACY;
       antiwindup_strat_.i_max = i_max;
       antiwindup_strat_.i_min = i_min;
+      antiwindup_strat_.legacy_antiwindup = false;
       antiwindup_strat_.validate();
     }
 
@@ -289,6 +290,7 @@ public:
       antiwindup_strat_.type = AntiwindupStrategy::LEGACY;
       antiwindup_strat_.i_max = i_max;
       antiwindup_strat_.i_min = i_min;
+      antiwindup_strat_.legacy_antiwindup = antiwindup;
       antiwindup_strat_.validate();
     }
 
@@ -314,6 +316,7 @@ public:
       i_min_(antiwindup_strat.i_min),
       u_max_(u_max),
       u_min_(u_min),
+      antiwindup_(antiwindup_strat.legacy_antiwindup),
       antiwindup_strat_(antiwindup_strat)
     {
       if (std::isnan(u_min) || std::isnan(u_max))
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -308,7 +308,8 @@ public:
    *
    */
     Gains(
-      double p, double i, double d, double u_max, double u_min, AntiwindupStrategy antiwindup_strat)
+      double p, double i, double d, double u_max, double u_min,
+      const AntiwindupStrategy & antiwindup_strat)
     : p_gain_(p),
       i_gain_(i),
       d_gain_(d),
@@ -397,7 +398,8 @@ public:
    * \throws An std::invalid_argument exception is thrown if u_min > u_max
    */
   Pid(
-    double p, double i, double d, double u_max, double u_min, AntiwindupStrategy antiwindup_strat);
+    double p, double i, double d, double u_max, double u_min,
+    const AntiwindupStrategy & antiwindup_strat);
 
   /*!
    * \brief Copy constructor required for preventing mutexes from being copied
@@ -446,7 +448,8 @@ public:
    * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
    */
   void initialize(
-    double p, double i, double d, double u_max, double u_min, AntiwindupStrategy antiwindup_strat);
+    double p, double i, double d, double u_max, double u_min,
+    const AntiwindupStrategy & antiwindup_strat);
 
   /*!
    * \brief Reset the state of this PID controller
@@ -570,7 +573,8 @@ public:
    * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
    */
   void set_gains(
-    double p, double i, double d, double u_max, double u_min, AntiwindupStrategy antiwindup_strat);
+    double p, double i, double d, double u_max, double u_min,
+    const AntiwindupStrategy & antiwindup_strat);
 
   /*!
    * \brief Set PID gains for the controller.
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -137,6 +137,12 @@ public:
       std::cerr << "AntiwindupStrategy 'legacy' requires i_min < i_max and to be finite."
                 << std::endl;
     }
+    if (
+      type != NONE && type != UNDEFINED && type != LEGACY && type != INTEGRATOR_CLAMPING &&
+      type != BACK_CALCULATION && type != CONDITIONAL_INTEGRATION)
+    {
+      throw std::runtime_error("AntiwindupStrategy has an invalid type.");
+    }
   }
 
   operator std::string() const { return to_string(); }
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -233,9 +233,6 @@ inline bool is_zero(T value, T tolerance = std::numeric_limits<T>::epsilon())
 
   \param u_clamp Minimum and maximum bounds for the controller output. The clamp is applied to the \f$command\f$.
 
-  \param trk_tc Tracking time constant for the 'back_calculation' strategy.
-
-
   \section Usage
 
   To use the Pid class, you should first call some version of init()
@@ -471,8 +468,6 @@ public:
    * \param d The derivative gain.
    * \param u_max Upper output clamp.
    * \param u_min Lower output clamp.
-   * \param trk_tc Specifies the tracking time constant for the 'back_calculation' strategy. If set
-   *    to 0.0 when this strategy is selected, a recommended default value will be applied.
    * \param antiwindup_strat Specifies the anti-windup strategy. Options: 'back_calculation',
         'conditional_integration', or 'none'. Note that the 'back_calculation' strategy use the
         tracking_time_constant parameter to tune the anti-windup behavior.
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -62,7 +62,7 @@ namespace control_toolbox
       integral contribution to the control output. i_max and
       i_min are applied in both scenarios.
  */
-struct AntiwindupStrategy
+struct AntiWindupStrategy
 {
 public:
   enum Value : int8_t
@@ -75,7 +75,7 @@ public:
     CONDITIONAL_INTEGRATION
   };
 
-  AntiwindupStrategy()
+  AntiWindupStrategy()
   : type(UNDEFINED),
     i_min(std::numeric_limits<double>::quiet_NaN()),
     i_max(std::numeric_limits<double>::quiet_NaN()),
@@ -117,12 +117,12 @@ public:
   {
     if (type == UNDEFINED)
     {
-      throw std::invalid_argument("AntiwindupStrategy is UNDEFINED. Please set a valid type");
+      throw std::invalid_argument("AntiWindupStrategy is UNDEFINED. Please set a valid type");
     }
     if (type == BACK_CALCULATION && (trk_tc < 0.0 || !std::isfinite(trk_tc)))
     {
       throw std::invalid_argument(
-        "AntiwindupStrategy 'back_calculation' requires a valid positive tracking time constant "
+        "AntiWindupStrategy 'back_calculation' requires a valid positive tracking time constant "
         "(trk_tc)");
     }
     if (
@@ -131,7 +131,7 @@ public:
     {
       throw std::invalid_argument(
         fmt::format(
-          "AntiwindupStrategy 'integrator_clamping' requires i_min < i_max and to be finite "
+          "AntiWindupStrategy 'integrator_clamping' requires i_min < i_max and to be finite "
           "(i_min: {}, i_max: {})",
           i_min, i_max));
     }
@@ -139,7 +139,7 @@ public:
     {
       throw std::invalid_argument(
         fmt::format(
-          "AntiwindupStrategy 'legacy' requires i_min < i_max and to be finite (i_min: {}, i_max: "
+          "AntiWindupStrategy 'legacy' requires i_min < i_max and to be finite (i_min: {}, i_max: "
           "{})",
           i_min, i_max));
     }
@@ -147,7 +147,7 @@ public:
       type != NONE && type != UNDEFINED && type != LEGACY && type != INTEGRATOR_CLAMPING &&
       type != BACK_CALCULATION && type != CONDITIONAL_INTEGRATION)
     {
-      throw std::invalid_argument("AntiwindupStrategy has an invalid type");
+      throw std::invalid_argument("AntiWindupStrategy has an invalid type");
     }
   }
 
@@ -279,7 +279,7 @@ public:
    * \param i_min Lower integral clamp.
    *
    */
-    [[deprecated("Use constructor with AntiwindupStrategy instead.")]]
+    [[deprecated("Use constructor with AntiWindupStrategy instead.")]]
     Gains(double p, double i, double d, double i_max, double i_min)
     : p_gain_(p),
       i_gain_(i),
@@ -290,7 +290,7 @@ public:
       u_min_(-std::numeric_limits<double>::infinity()),
       antiwindup_(false)
     {
-      antiwindup_strat_.type = AntiwindupStrategy::LEGACY;
+      antiwindup_strat_.type = AntiWindupStrategy::LEGACY;
       antiwindup_strat_.i_max = i_max;
       antiwindup_strat_.i_min = i_min;
       antiwindup_strat_.legacy_antiwindup = false;
@@ -310,7 +310,7 @@ public:
         i_min are applied in both scenarios.
    *
    */
-    [[deprecated("Use constructor with AntiwindupStrategy instead.")]]
+    [[deprecated("Use constructor with AntiWindupStrategy instead.")]]
     Gains(double p, double i, double d, double i_max, double i_min, bool antiwindup)
     : p_gain_(p),
       i_gain_(i),
@@ -321,7 +321,7 @@ public:
       u_min_(-std::numeric_limits<double>::infinity()),
       antiwindup_(antiwindup)
     {
-      antiwindup_strat_.type = AntiwindupStrategy::LEGACY;
+      antiwindup_strat_.type = AntiWindupStrategy::LEGACY;
       antiwindup_strat_.i_max = i_max;
       antiwindup_strat_.i_min = i_min;
       antiwindup_strat_.legacy_antiwindup = antiwindup;
@@ -342,7 +342,7 @@ public:
    */
     Gains(
       double p, double i, double d, double u_max, double u_min,
-      const AntiwindupStrategy & antiwindup_strat)
+      const AntiWindupStrategy & antiwindup_strat)
     : p_gain_(p),
       i_gain_(i),
       d_gain_(d),
@@ -383,7 +383,7 @@ public:
         error_msg = "Gains: u_min or u_max must not be NaN";
         return false;
       }
-      else if (antiwindup_strat_.type == AntiwindupStrategy::UNDEFINED)
+      else if (antiwindup_strat_.type == AntiWindupStrategy::UNDEFINED)
       {
         error_msg =
           "Gains: Antiwindup strategy cannot be UNDEFINED. Please set a valid antiwindup strategy.";
@@ -403,7 +403,7 @@ public:
 
     // Default constructor
     [[deprecated(
-      "Use constructor with AntiwindupStrategy only. The default constructor might be deleted in "
+      "Use constructor with AntiWindupStrategy only. The default constructor might be deleted in "
       "future")]] Gains()
     {
     }
@@ -427,7 +427,7 @@ public:
     double u_min_ = -std::numeric_limits<double>::infinity(); /**< Minimum allowable output. */
     [[deprecated("Use antiwindup_strat_ instead.")]]
     bool antiwindup_ = false;             /**< Anti-windup. */
-    AntiwindupStrategy antiwindup_strat_; /**< Anti-windup strategy. */
+    AntiWindupStrategy antiwindup_strat_; /**< Anti-windup strategy. */
   };
 
   /*!
@@ -446,7 +446,7 @@ public:
    *
    * \throws An std::invalid_argument exception is thrown if i_min > i_max
    */
-  [[deprecated("Use constructor with AntiwindupStrategy only.")]]
+  [[deprecated("Use constructor with AntiWindupStrategy only.")]]
   Pid(
     double p = 0.0, double i = 0.0, double d = 0.0, double i_max = 0.0, double i_min = -0.0,
     bool antiwindup = false);
@@ -467,7 +467,7 @@ public:
    */
   Pid(
     double p, double i, double d, double u_max, double u_min,
-    const AntiwindupStrategy & antiwindup_strat);
+    const AntiWindupStrategy & antiwindup_strat);
 
   /*!
    * \brief Copy constructor required for preventing mutexes from being copied
@@ -495,7 +495,7 @@ public:
    * \return True if all parameters are successfully set, False otherwise.
    * \note New gains are not applied if i_min_ > i_max_
    */
-  [[deprecated("Use initialize with AntiwindupStrategy instead.")]]
+  [[deprecated("Use initialize with AntiWindupStrategy instead.")]]
   bool initialize(
     double p, double i, double d, double i_max, double i_min, bool antiwindup = false);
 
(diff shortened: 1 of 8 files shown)


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,138 @@
...
+    }
+    else if (s == "none")
+    {
+      type = NONE;
     }
     else
     {
-      value_ = NONE;
+      type = UNDEFINED;
+      std::cerr << "Unknown antiwindup strategy: '" << s << "'. Using UNDEFINED." << std::endl;
```
fmauch: Wouldn't it make sense to even throw an exception here? I see that an exception is thrown in the `validate` function, though.

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,138 @@
...
     }
   }
 
-private:
-  Value value_;
+  Value type = UNDEFINED;
+  double i_min = std::numeric_limits<double>::quiet_NaN(); /**< Minimum allowable integral term. */
+  double i_max = std::numeric_limits<double>::quiet_NaN(); /**< Maximum allowable integral term. */
+
+  bool legacy_antiwindup = false; /**< Use legacy anti-windup strategy. */
```
fmauch, line 188: Is that comment describing this correctly? From the code below it seems to me, the meaning of this member is more something like "Enable anti-windup if the LEGACY windup strategy is being used"

control_toolbox/src/pid.cpp
```diff
@@ -166,134 +148,114 @@ void Pid::get_gains(double & p, double & i, double & d, double & i_max, double &
...
+  return false;
+}
+
+bool Pid::set_gains(const Gains & gains_in)
+{
+  std::string error_msg = "";
+  if (!gains_in.validate(error_msg))
   {
-    std::cout << "Received NaN for u_min or u_max, skipping new gains" << std::endl;
+    std::cout << "PID: Invalid gains: " << error_msg << ". SKipping new gains." << std::endl;
```
fmauch: The other error outputs above have gone to `cerr`. Is there a reason why this one is not?

control_toolbox/src/pid.cpp
```diff
@@ -383,19 +345,30 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
     // Prevent i_term_ from climbing higher than permitted by i_max_/i_min_
     i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_, gains.i_min_, gains.i_max_);
   }
-  else if (!gains.antiwindup_ && gains.antiwindup_strat_ == AntiwindupStrategy::NONE)
+  else if (
+    !gains.antiwindup_strat_.legacy_antiwindup &&
+    gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
   {
     i_term_ += gains.i_gain_ * dt_s * p_error_;
   }
```
fmauch: Suggestion: Restructure to a nested if

```C++
if (gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
{
  if (gains.antiwindup_strat_.legacy_antiwindup)
  {
    i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_, gains.i_min_, gains.i_max_);
  }
  else
  {
    i_term_ += gains.i_gain_ * dt_s * p_error_;
  }
}

File: control_toolbox/src/pid_ros.cpp [hunk 5]
```diff
@@ -241,21 +242,36 @@ bool PidROS::initialize_from_ros_parameters()
     set_parameter_event_callback();
   }
 
-  if (antiwindup_strat_str == "none")
+  if (antiwindup_strat_str == "legacy")
   {
     std::cout << "Old anti-windup technique is deprecated. "
                  "This option will be removed by the ROS 2 Kilted Kaiju release."
```
fmauch: Is it still planned to be removed for Kilted?

control_toolbox/src/pid_ros.cpp
```diff
@@ -656,17 +616,18 @@ void PidROS::set_parameter_event_callback()
...
         {
-          gains.antiwindup_ = parameter.get_value<bool>();
+          gains.antiwindup_strat_.legacy_antiwindup = parameter.get_value<bool>();
           changed = true;
         }
         else if (param_name == param_prefix_ + "antiwindup_strategy")
         {
-          gains.antiwindup_strat_ = AntiwindupStrategy(parameter.get_value<std::string>());
+          // @todo decide if this can be changed in the first place
+          // gains.antiwindup_strat_ = AntiWindupStrategy(parameter.get_value<std::string>());
```
fmauch: What do you mean by that?

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Not really, The LEGACY mode will be removed in the future, and so this one `legacy_antiwindup` option, and this was added to add support to how things were working in the past. So, no behavior change

control_toolbox/src/pid.cpp
saikishor: Good point.

control_toolbox/src/pid_ros.cpp
saikishor: Probably yes. @christophfroehlich am I right?

control_toolbox/src/pid_ros.cpp
saikishor: We have to decide on the fact that the antiwindup strategy can be changed on-the-fly or it should be a read-only argument

control_toolbox/src/pid.cpp
saikishor: Done

control_toolbox/src/pid.cpp
saikishor: Done

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Done

control_toolbox/src/pid_ros.cpp
christophfroehlich: let's discuss this in the next meeting. We have too many unfinished cleanup tasks which should have been gone into Kilted. And I have the feeling that our Kilted packages are still in "development" and we don't claim them to be stable.

control_toolbox/src/pid_ros.cpp
christophfroehlich: I don't think we should allow this, I don't see a use-case for doing so.

control_toolbox/src/pid.cpp
```diff
@@ -166,134 +148,114 @@ void Pid::get_gains(double & p, double & i, double & d, double & i_max, double &
...
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-void Pid::set_gains(
-  double p, double i, double d, double i_max, double i_min, double u_max, double u_min,
-  double trk_tc, bool antiwindup, AntiwindupStrategy antiwindup_strat)
-{
-  Gains gains(p, i, d, i_max, i_min, u_max, u_min, trk_tc, antiwindup, antiwindup_strat);
-
-  set_gains(gains);
+    Gains gains(p, i, d, i_max, i_min, antiwindup);
+#pragma GCC diagnostic pop
```
christophfroehlich: we have nested/double pragma GCC here.

control_toolbox/src/pid_ros.cpp
saikishor: Thanks

saikishor: I just removed it

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
+ * \param tracking_time_constant Specifies the tracking time constant for the 'back_calculation' strategy. If set
+ *    to 0.0 when this strategy is selected, a recommended default value will be applied.
+ * \param legacy_antiwindup Anti-windup functionality. When set to true, limits
+      the integral error to prevent windup; otherwise, constrains the
+      integral contribution to the control output. i_max and
+      i_min are applied in both scenarios.
+  * \param error_deadband Error deadband is used to avoid integration when the error is within a small range.
+  * \param type Specifies the antiwindup strategy type. Valid values are:
+ *   - `NONE`: No antiwindup strategy applied.
+ *   - `LEGACY`: Legacy antiwindup strategy, which limits the integral term to prevent windup.
```
```suggestion
 *   - `LEGACY`: Legacy antiwindup strategy, which limits the integral term to prevent windup.
```
```suggestion
 *   - `LEGACY`: Legacy antiwindup strategy, which limits the integral term to prevent windup (deprecated: This option will be removed in a future release).
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
+    UNDEFINED = -1,
+    NONE,
+    LEGACY,
     BACK_CALCULATION,
     CONDITIONAL_INTEGRATION
   };
 
-  constexpr AntiwindupStrategy() : value_(NONE) {}
-  constexpr AntiwindupStrategy(Value v) : value_(v) {}  // NOLINT(runtime/explicit)
+  AntiWindupStrategy()
```
christophfroehlich, line 83: Should we add a constructor with parameters? Just a note, I'm fine without it.

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
+ * \param i_min Lower integral clamp.
+ * \param u_max Upper output clamp.
+ * \param u_min Lower output clamp.
+ * \param tracking_time_constant Specifies the tracking time constant for the 'back_calculation' strategy. If set
+ *    to 0.0 when this strategy is selected, a recommended default value will be applied.
+ * \param legacy_antiwindup Anti-windup functionality. When set to true, limits
+      the integral error to prevent windup; otherwise, constrains the
+      integral contribution to the control output. i_max and
+      i_min are applied in both scenarios.
+  * \param error_deadband Error deadband is used to avoid integration when the error is within a small range.
```
```suggestion
  * \param error_deadband Error deadband is used to avoid integration when the error is within a small range.
```
```suggestion
  * \param error_deadband Error deadband is used to stop integration when the error is within the given range.
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
+      type = BACK_CALCULATION;
     }
     else if (s == "conditional_integration")
     {
-      value_ = CONDITIONAL_INTEGRATION;
+      type = CONDITIONAL_INTEGRATION;
+    }
+    else if (s == "legacy")
+    {
+      type = LEGACY;
```
christophfroehlich, line 105: should we add a deprecation notice here?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
+      type = UNDEFINED;
+      throw std::invalid_argument(
+        "AntiWindupStrategy: Unknown antiwindup strategy : '" + s +
+        "'. Valid strategies are: 'back_calculation', 'conditional_integration', 'legacy', "
+        "'none'.");
     }
   }
 
-  operator std::string() const { return to_string(); }
+  void validate() const
```
christophfroehlich, line 124: should we add a deprecation notice here for
`std::isfinite(i_min) || std::isfinite(i_max)`, independent of the type?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -39,64 +39,151 @@
...
-  Value value_;
+  Value type = UNDEFINED;
+  double i_min = std::numeric_limits<double>::quiet_NaN(); /**< Minimum allowable integral term. */
+  double i_max = std::numeric_limits<double>::quiet_NaN(); /**< Maximum allowable integral term. */
+
+  bool legacy_antiwindup = false; /**< Use legacy anti-windup strategy. */
+
+  // tracking_time_constant Specifies the tracking time constant for the 'back_calculation'
+  // strategy. If set to 0.0 when this strategy is selected, a recommended default value
+  // will be applied.
```
```suggestion
  // strategy. If set to 0.0 when this strategy is selected, a recommended default value
  // will be applied.
```
```suggestion
  // strategy. If set to 0.0 a recommended default value will be applied.
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,31 +365,44 @@ class Pid
...
+      {
+        error_msg = "Gains: u_min or u_max must not be NaN";
+        return false;
+      }
+      else if (antiwindup_strat_.type == AntiWindupStrategy::UNDEFINED)
+      {
+        error_msg =
+          "Gains: Antiwindup strategy cannot be UNDEFINED. Please set a valid antiwindup strategy.";
+        return false;
+      }
```
christophfroehlich: is this necessary here? this is already checked in the AntiWindupStrategy?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -307,21 +411,22 @@ class Pid
...
 
     double p_gain_ = 0.0; /**< Proportional gain. */
     double i_gain_ = 0.0; /**< Integral gain. */
     double d_gain_ = 0.0; /**< Derivative gain. */
-    double i_max_ = 0.0;  /**< Maximum allowable integral term. */
-    double i_min_ = 0.0;  /**< Minimum allowable integral term. */
+    [[deprecated("Use antiwindup_strat_.i_max instead.")]]
+    double i_max_ = 0.0; /**< Maximum allowable integral term. */
+    [[deprecated("Use antiwindup_strat_.i_min instead.")]]
+    double i_min_ = 0.0; /**< Minimum allowable integral term. */
```
```suggestion
    [[deprecated("Use antiwindup_strat_.i_max instead.")]]
    double i_max_ = 0.0; /**< Maximum allowable integral term. */
    [[deprecated("Use antiwindup_strat_.i_min instead.")]]
    double i_min_ = 0.0; /**< Minimum allowable integral term. */
```
```suggestion
    double i_max_ = 0.0; /**< Maximum allowable integral term. */
    double i_min_ = 0.0; /**< Minimum allowable integral term. */
```
we get tons of deprecation warnings from our library itself

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -307,21 +411,22 @@ class Pid
...
     double u_max_ = std::numeric_limits<double>::infinity();  /**< Maximum allowable output. */
     double u_min_ = -std::numeric_limits<double>::infinity(); /**< Minimum allowable output. */
-    double trk_tc_ = 0.0;                                     /**< Tracking time constant. */
-    bool antiwindup_ = false;                                 /**< Anti-windup. */
-    AntiwindupStrategy antiwindup_strat_ = AntiwindupStrategy::NONE; /**< Anti-windup strategy. */
+    [[deprecated("Use antiwindup_strat_ instead.")]]
+    bool antiwindup_ = false;             /**< Anti-windup. */
```
```suggestion
    [[deprecated("Use antiwindup_strat_ instead.")]]
    bool antiwindup_ = false;             /**< Anti-windup. */
```
```suggestion
    bool antiwindup_ = false;             /**< Anti-windup. */
```

we get tons of deprecation warnings from our library itself

control_toolbox/test/pid_tests.cpp
```diff
@@ -426,34 +441,38 @@ TEST(ParameterTest, gainSettingCopyPIDTest)
...
   // Initialize the default way
-  Pid pid1(
-    p_gain, i_gain, d_gain, i_max, i_min, u_max, u_min, trk_tc, antiwindup, antiwindup_strat);
+  Pid pid1(p_gain, i_gain, d_gain, u_max, u_min, antiwindup_strat);
 
   // Test return values  -------------------------------------------------
   double p_gain_return, i_gain_return, d_gain_return, i_max_return, i_min_return, u_max_return,
-    u_min_return, trk_tc_return;
+    u_min_return, tracking_time_constant_return;
   bool antiwindup_return;
```
christophfroehlich:
```
/workspaces/ros2_rolling_ws/src/control_toolbox/control_toolbox/test/pid_tests.cpp:457:55: warning: unused variable ‘i_max_return’ [-Wunused-variable]
  457 |   double p_gain_return, i_gain_return, d_gain_return, i_max_return, i_min_return, u_max_return,
      |                                                       ^~~~~~~~~~~~
/workspaces/ros2_rolling_ws/src/control_toolbox/control_toolbox/test/pid_tests.cpp:457:69: warning: unused variable ‘i_min_return’ [-Wunused-variable]
  457 |   double p_gain_return, i_gain_return, d_gain_return, i_max_return, i_min_return, u_max_return,
      |                                                                     ^~~~~~~~~~~~
/workspaces/ros2_rolling_ws/src/control_toolbox/control_toolbox/test/pid_tests.cpp:458:19: warning: unused variable ‘tracking_time_constant_return’ [-Wunused-variable]
  458 |     u_min_return, tracking_time_constant_return;
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/workspaces/ros2_rolling_ws/src/control_toolbox/control_toolbox/test/pid_tests.cpp:459:8: warning: unused variable ‘antiwindup_return’ [-Wunused-variable]
  459 |   bool antiwindup_return;
      |        ^~~~~~~~~~~~~~~~~
---
```

control_toolbox/include/control_toolbox/pid.hpp
saikishor: So, better no to mention the deprecation. Fair enough

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -194,12 +278,21 @@ class Pid
...
+      i_max_(i_max),
+      i_min_(i_min),
+      u_max_(std::numeric_limits<double>::infinity()),
+      u_min_(-std::numeric_limits<double>::infinity()),
+      antiwindup_(false)
     {
+      antiwindup_strat_.type = AntiWindupStrategy::LEGACY;
+      antiwindup_strat_.i_max = i_max;
+      antiwindup_strat_.i_min = i_min;
+      antiwindup_strat_.legacy_antiwindup = false;
```
```suggestion
      antiwindup_strat_.legacy_antiwindup = true;
```
wasn't true here the default?

control_toolbox/include/control_toolbox/pid.hpp
saikishor: If needed we can always add in future, but for now I think it is better like this

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Only one at the beginning is enough, We don't have to print everytime the user changes the gains. What do you think?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -515,47 +561,20 @@ class Pid
   void get_gains(
```
christophfroehlich, line 561: we should also deprecate (we want to remove i_min/i_max)
`void get_gains(double & p, double & i, double & d, double & i_max, double & i_min);`
and change line
`[[deprecated("Use get_gains overload without bool antiwindup.")]]`
to 
`[[deprecated("Use get_gains overload with AntiWindupStrategy argument.")]]`

control_toolbox/src/pid.cpp
saikishor: Done

control_toolbox/src/pid.cpp
```diff
@@ -50,46 +50,36 @@ constexpr double UMAX_INFINITY = std::numeric_limits<double>::infinity();
...
   }
-  else if (u_min > u_max)
-  {
-    throw std::invalid_argument("received u_min > u_max");
-  }
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-  set_gains(p, i, d, i_max, i_min, u_max, u_min, trk_tc, antiwindup, antiwindup_strat);
-#pragma GCC diagnostic pop
+  AntiWindupStrategy antiwindup_strat;
```
christophfroehlich, line 59: we can remove the i_min check here, as we check it later in 
```
bool Pid::set_gains(
  double p, double i, double d, double u_max, double u_min,
  const AntiWindupStrategy & antiwindup_strat)
```

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Done

control_toolbox/include/control_toolbox/pid.hpp
saikishor: You are right. Removed

control_toolbox/test/pid_tests.cpp
saikishor: Done. Removed

control_toolbox/src/pid.cpp
```diff
@@ -383,19 +339,32 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
-    i_term_ = std::clamp(i_term_ + gains.i_gain_ * dt_s * p_error_, gains.i_min_, gains.i_max_);
+    throw std::runtime_error(
+      "PID: Antiwindup strategy cannot be UNDEFINED. Please set a valid antiwindup strategy.");
   }
-  else if (!gains.antiwindup_ && gains.antiwindup_strat_ == AntiwindupStrategy::NONE)
+
+  // Calculate integral contribution to command
+  const bool is_error_in_deadband_zone =
+    control_toolbox::is_zero(error, gains.antiwindup_strat_.error_deadband);
+  if (!is_error_in_deadband_zone && gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
```
christophfroehlich, line 351:
```suggestion
  const bool is_error_in_deadband_zone =
    control_toolbox::is_zero(error, gains.antiwindup_strat_.error_deadband);
  if (gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
```
should we really change the behavior here for the LEGACY?

control_toolbox/src/pid.cpp
```diff
@@ -383,19 +339,32 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
+    {
+      i_term_ += gains.i_gain_ * dt_s * p_error_;
+    }
   }
 
   // Compute the command
-  if (!gains.antiwindup_ && gains.antiwindup_strat_ == AntiwindupStrategy::NONE)
+  if (
+    !is_error_in_deadband_zone && !gains.antiwindup_strat_.legacy_antiwindup &&
+    gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
```
christophfroehlich:
```suggestion
  if (
    !gains.antiwindup_strat_.legacy_antiwindup &&
    gains.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
```

control_toolbox/src/pid_ros.cpp
```diff
@@ -241,21 +244,35 @@ bool PidROS::initialize_from_ros_parameters()
...
-    std::cout << "Old anti-windup technique is deprecated. "
-                 "This option will be removed by the ROS 2 Kilted Kaiju release."
-              << std::endl;
+    antiwindup_strat.validate();
+  }
+  catch (const std::exception & e)
+  {
+    RCLCPP_ERROR(node_logging_->get_logger(), "Invalid antiwindup strategy: %s", e.what());
+    return false;
   }
```
christophfroehlich, line 268: can't we skip this check here, as it will be done anyways in 
```
bool Pid::set_gains(
  double p, double i, double d, double u_max, double u_min,
  const AntiWindupStrategy & antiwindup_strat)
```

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: I mean, especially if the type is not LEGACY, then i_min and i_max should be NaN (they have no effect). At least I can't find any deprecation warning then.

control_toolbox/src/pid.cpp
saikishor: By default, it is `epsilon`, so mostly we are not changing the behavior unless someone explicitly sets it

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Makes sense. Sorry I missed it 

https://github.com/ros-controls/control_toolbox/blob/049962ac327f679035db1714a464e428638fb14f/control_toolbox/include/control_toolbox/pid.hpp#L127-L130

control_toolbox/src/pid.cpp
christophfroehlich: actually, I wanted to write if we want to enable to change the LEGACY behavior, as we want to remove it anyways. but I'm fine with keeping it

control_toolbox/src/pid.cpp
saikishor: I had to leave that for backward compatibility reasons. Because old method used to throw an exception and now we check but don't throw exceptions. That's why

control_toolbox/src/pid.cpp
saikishor: ok 👍🏾

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Got it

control_toolbox/src/pid_ros.cpp
saikishor: Done

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Done

christophfroehlich (dismissed):
Thanks for the nice work!

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,31 +374,38 @@ class Pid
...
-        tracking_time_constant parameter to tune the anti-windup behavior.
-   *
-   */
-    Gains(
-      double p, double i, double d, double u_max, double u_min, double trk_tc,
-      AntiwindupStrategy antiwindup_strat)
-    : Gains(p, i, d, 0.0, 0.0, u_max, u_min, trk_tc, false, antiwindup_strat)
+    bool validate(std::string & error_msg) const
     {
+      if (i_min_ > i_max_)
```
fmauch, line 379: The error message says it has to be smaller, but the check is smaller or equal.
```suggestion
      if (i_min_ >= i_max_)
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,31 +374,38 @@ class Pid
...
+      {
+        error_msg = fmt::format("Gains: i_min ({}) must be less than i_max ({})", i_min_, i_max_);
+        return false;
+      }
+      else if (u_min_ > u_max_)
```
fmauch:
```suggestion
      else if (u_min_ >= u_max_)
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -274,31 +374,38 @@ class Pid
...
+      {
+        error_msg = fmt::format("Gains: u_min ({}) must be less than u_max ({})", u_min_, u_max_);
+        return false;
+      }
+      else if (std::isnan(u_min_) || std::isnan(u_max_))
+      {
+        error_msg = "Gains: u_min or u_max must not be NaN";
```
fmauch:
```suggestion
        error_msg = "Gains: u_min and u_max must not be NaN";
```

christophfroehlich (dismissed):

urfeex:
Some tests are failing, though.

saikishor:
> Some tests are failing, though.

I'll check

control_toolbox/include/control_toolbox/pid.hpp
saikishor: I reverted this change alone to not break the behavioural change. We can update it back in a new PR for rolling and kilted.

christophfroehlich (approved):

PR #424, opened 2025-07-01
Add CI job for docs

christophfroehlich:
https://github.com/ros-controls/control.ros.org/pull/470

### DIFF
--- a/.github/workflows/jazzy-check-docs.yml
+++ b/.github/workflows/jazzy-check-docs.yml
@@ -8,7 +8,14 @@ on:
     paths:
       - '**.rst'
       - '**.md'
+      - '**.jpg'
+      - '**.jpeg'
+      - '**.png'
+      - '**.svg'
+      - '**.yml'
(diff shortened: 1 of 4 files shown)


saikishor (approved):

Raivias (commented):
To make this workflow more uniform and robust to copy into other repos I'd like to suggest adding the following file formats to the list.
- .jpg
- .jpeg
- .png
- .svg
- .yml
- .bib maybe

christophfroehlich:
> To make this workflow more uniform and robust to copy into other repos I'd like to suggest adding the following file formats to the list.
> 
>     * .jpg
> 
>     * .jpeg
> 
>     * .png
> 
>     * .svg
> 
>     * .yml
> 
>     * .bib maybe

You mean for the path filters? Could make sense. problem with the yml files is that the github workflows also have the same ending. I can have a look if one can exclude the .github folder but include other yaml files.

PR #431, opened 2025-07-20
Deprecate `prefix_is_for_params` of PidROS

### LINKED ISSUE #156: Remove `prefix_is_for_params` from PID_Ros
https://github.com/ros-controls/control_toolbox/pull/129#pullrequestreview-1371661251

christophfroehlich:
See https://github.com/ros-controls/control_toolbox/issues/156#issuecomment-2985605034

Fixes #156

### DIFF
--- a/control_toolbox/include/control_toolbox/pid_ros.hpp
+++ b/control_toolbox/include/control_toolbox/pid_ros.hpp
@@ -40,7 +40,6 @@
 
 #include "rclcpp/clock.hpp"
 #include "rclcpp/duration.hpp"
-#include "rclcpp/logging.hpp"  // TODO(christophfroehlich) remove after deprecation removal
 #include "rclcpp/node.hpp"
 
 #include "realtime_tools/realtime_buffer.hpp"
@@ -77,11 +76,6 @@ public:
       node_ptr->get_node_parameters_interface(), node_ptr->get_node_topics_interface(), prefix,
       prefix_is_for_params)
   {
-    // note: deprecation on templated constructor does not show up
-    RCLCPP_INFO(
-      node_ptr->get_logger(),
-      "PidROS constructor with node and prefix is deprecated, use overloads with explicit "
-      "prefixes for params and topics");
   }
   template <class NodeT>
   explicit PidROS(std::shared_ptr<NodeT> node_ptr, const std::string & param_prefix)
--- a/control_toolbox/include/control_toolbox/pid_ros.hpp
+++ b/control_toolbox/include/control_toolbox/pid_ros.hpp
@@ -85,6 +85,17 @@ public:
       param_prefix, std::string(""), false)
   {
   }
+  /*!
+   * \brief Constructor of PidROS class.
+   *
+   * The node is passed to this class to handler the ROS parameters, this class allows
+   * to add a prefix to the pid parameters
+   *
+   * \param node Any ROS node
+   * \param param_prefix prefix to add to the pid parameters. This string is not manipulated, i.e., probably should end with `.`
+   * \param topic_prefix prefix to add to the state publisher. This string is not manipulated, i.e., probably should end with `/`. If it starts with `~/`, topic will be local under the namespace of the node. If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   *
+   */
   template <class NodeT>
   explicit PidROS(
     std::shared_ptr<NodeT> node_ptr, std::string param_prefix, std::string topic_prefix)
@@ -94,6 +105,18 @@ public:
       param_prefix, topic_prefix, true)
   {
   }
+  /*!
+   * \brief Constructor of PidROS class.
+   *
+   * The node is passed to this class to handler the ROS parameters, this class allows
+   * to add a prefix to the pid parameters
+   *
+   * \param node Any ROS node
+   * \param param_prefix prefix to add to the pid parameters. This string is not manipulated, i.e., probably should end with `.`
+   * \param topic_prefix prefix to add to the state publisher. This string is not manipulated, i.e., probably should end with `/`. If it starts with `~/`, topic will be local under the namespace of the node. If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   * \param activate_state_publisher If true, the publisher will be enabled after initialization.
+   *
+   */
   template <class NodeT>
   explicit PidROS(
     std::shared_ptr<NodeT> node_ptr, std::string param_prefix, std::string topic_prefix,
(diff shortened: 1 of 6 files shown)


control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -77,14 +77,82 @@ class PidROS
       prefix_is_for_params)
   {
   }
+  template <class NodeT>
+  explicit PidROS(std::shared_ptr<NodeT> node_ptr, std::string param_prefix = std::string(""))
```
saikishor:
```suggestion
  explicit PidROS(std::shared_ptr<NodeT> node_ptr, const std::string & param_prefix)
```

Better no default parameters at all

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -77,14 +77,82 @@ class PidROS
...
+   * to add a prefix to the pid parameters
+   *
+   * \param node Any ROS node
+   * \param param_prefix prefix to add to the pid parameters. This string is not manipulated, i.e., probably should end with `.`
+   * \param topic_prefix prefix to add to the state publisher. This string is not manipulated, i.e., probably should end with `/`. If it starts with `~/`, topic will be local under the namespace of the node. If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   *
+   */
+  template <class NodeT>
+  explicit PidROS(
+    std::shared_ptr<NodeT> node_ptr, std::string param_prefix, std::string topic_prefix)
```
saikishor:
```suggestion
    std::shared_ptr<NodeT> node_ptr, const std::string &param_prefix, const std::string &topic_prefix)
```

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -77,14 +77,82 @@ class PidROS
       prefix_is_for_params)
   {
   }
+  template <class NodeT>
+  explicit PidROS(std::shared_ptr<NodeT> node_ptr, std::string param_prefix = std::string(""))
+  : PidROS(
+      node_ptr->get_node_base_interface(), node_ptr->get_node_logging_interface(),
+      node_ptr->get_node_parameters_interface(), node_ptr->get_node_topics_interface(),
+      param_prefix, std::string(""), false)
```
saikishor:
```suggestion
      param_prefix, "", false)
```

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -77,14 +77,82 @@ class PidROS
...
+   * \param param_prefix Prefix to add to the PID parameters. This string is not manipulated, i.e., probably should end with `.`.
+   * \param topic_prefix Prefix to add to the state publisher. This string is not manipulated, i.e., probably should end with `/`. If it starts with `~/`, topic will be local under the namespace of the node. If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   * \param activate_state_publisher If true, the publisher will be enabled after initialization.
+   */
+  PidROS(
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base,
+    rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr node_logging,
+    rclcpp::node_interfaces::NodeParametersInterface::SharedPtr node_params,
+    rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr topics_interface,
+    std::string param_prefix, std::string topic_prefix, bool activate_state_publisher);
```
saikishor:
```suggestion
    const std::string &param_prefix, const std::string &topic_prefix, bool activate_state_publisher);
```

control_toolbox/src/pid_ros.cpp
```diff
@@ -632,6 +656,25 @@ void PidROS::set_parameter_event_callback()
...
+            RCLCPP_INFO(
+              node_logging_->get_logger(), "Activate publisher: `%s` ...", topic_name.c_str());
+            state_pub_ = rclcpp::create_publisher<control_msgs::msg::PidState>(
+              topics_interface_, topic_name, rclcpp::SensorDataQoS());
+            rt_state_pub_.reset(
+              new realtime_tools::RealtimePublisher<control_msgs::msg::PidState>(state_pub_));
+          }
+          else
+          {
+            RCLCPP_INFO(node_logging_->get_logger(), "Deactivate publisher...");
```
saikishor, line 690:
```suggestion
            RCLCPP_INFO(node_logging_->get_logger(), "Not publishing any state information...");
```

Deactivate doesn't sound right here. What do you think?

control_toolbox/src/pid_ros.cpp
```diff
@@ -352,6 +373,9 @@ bool PidROS::initialize_from_args(
         param_prefix_ + "antiwindup_strategy",
         rclcpp::ParameterValue(gains.antiwindup_strat_.to_string()));
       declare_param(param_prefix_ + "save_i_term", rclcpp::ParameterValue(save_i_term));
+      declare_param(
+        param_prefix_ + "activate_state_publisher",
```
saikishor, line 394:
```suggestion
        param_prefix_ + "publish_state",
```

Something simple like this?

control_toolbox/src/pid_ros.cpp
```diff
@@ -632,6 +656,25 @@ void PidROS::set_parameter_event_callback()
           gains.antiwindup_strat_.error_deadband = parameter.get_value<double>();
           changed = true;
         }
+        else if (param_name == param_prefix_ + "activate_state_publisher")
```
saikishor, line 676:
```suggestion
        else if (param_name == param_prefix_ + "publish_state")
```

control_toolbox/src/pid_ros.cpp
saikishor: As per our discussion over DM, it makes sense to call it deactivating as we are resetting the publisher

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -312,7 +382,7 @@ class PidROS
    *               If not stated explicitly using "/" or "~", prefix is interpreted as global, i.e.,
    *               "/" will be added in front of topic prefix
    */
-  void set_prefixes(const std::string & topic_prefix);
+  [[deprecated]] void set_prefixes(const std::string & topic_prefix);
```
destogl, line 385: We are removing this setter? Is this used anywhere in our codebase? I would be surprised if it is.

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -638,8 +630,8 @@ TEST(PidParametersTest, MultiplePidInstances)
 {
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("multiple_pid_instances");
 
-  TestablePidROS pid_1(node, "PID_1");
-  TestablePidROS pid_2(node, "PID_2");
+  TestablePidROS pid_1(node, "PID_1.");
```
destogl: Should we add "." Automatically for param prefixes? And "/" for topics?

destogl (changes requested):
Small clarification on default behavior. Otherwise good!

control_toolbox/include/control_toolbox/pid_ros.hpp
christophfroehlich: It is only used in the deprecated constructor, that's why I'd remove it

saikishor (dismissed):
LGTM

control_toolbox/src/pid_ros.cpp
destogl: Yes this is good!

destogl (approved):
Thanks! Let's also accept the @saikishor's proposal

control_toolbox/src/pid_ros.cpp
christophfroehlich: My idea was that `activate_state_publisher` is more descriptive because we essentially add/remove the publisher from the node, and don't only stop publishing. Does that make sense?

saikishor: Sure

control_toolbox/src/pid_ros.cpp
```diff
@@ -54,6 +54,12 @@ PidROS::PidROS(
   node_params_(node_params),
   topics_interface_(topics_interface)
 {
+  // note: deprecation on templated constructor does not show up
+  RCLCPP_INFO(
```
saikishor:
```suggestion
  RCLCPP_WARN(
```

Shouldn't it be a warning?

saikishor (approved):

PR #436, opened 2025-07-29
Remove legacy and deprecated PID parameters

ViktorCVS:
### **Overview**

This PR removes legacy and deprecated PID parameters from the `control_toolbox` package.

- Removed the legacy anti‑windup strategy
- Removed deprecated overloads

### **About tests**

The packages compile correctly and have passed the *pre‑commit* and *colcon* tests.

### **About change in tests**

Three test files were modified. In addition to the updates required by the removal of the legacy anti‑windup strategy, some tests were deleted and some assertions were swapped, so I think it is relevant to list them here.

-  pid_tests.cpp:

  Tests *ITermBadIBoundsTestConstructor*, *ITermBadIBoundsTest*, *integrationClampTest*, *integrationClampZeroGainTest*, *integrationAntiwindupTest* and *negativeIntegrationAntiwindupTest* were removed because they covered only the legacy anti‑windup strategy.

- pid_ros_parameters_tests.cpp:

  Because the legacy anti‑windup strategy was the only feature that caused `initialize_from_ros_parameters()` to return `false` with the default values, the expectation was changed to `ASSERT_TRUE`.

https://github.com/ros-controls/control_toolbox/commit/4b87edd93058d70242dd5a34ad328219bda02d74#diff-9779d5f0aefcf9960e4c7c562830de541fdfc64076a9c42ba32ab98e52dfdc43R466

  Test *InitPidTestBadParameter* was removed because it covered only the legacy anti‑windup strategy.

### **Related PR's**

- soon

### **Final notes**


control_toolbox/src/pid.cpp
```diff
@@ -357,32 +280,15 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
-      // Prevent i_term_ from climbing higher than permitted by i_max_/i_min_
-      i_term_ =
-        std::clamp(i_term_ + gains_.i_gain_ * dt_s * p_error_, gains_.i_min_, gains_.i_max_);
-    }
-    else
-    {
-      i_term_ += gains_.i_gain_ * dt_s * p_error_;
-    }
+    // No anti-windup strategy, so just integrate the error
+    i_term_ += dt_s * gains_.i_gain_ * error;
```
christophfroehlich: This is changing the behavior to my fix from #432. Both options are fine for me, I just thought I add it after calculating cmd_unsat_, at the same place as the other strategies are summing up i_term. Why you'd change it?

control_toolbox/test/pid_tests.cpp
```diff
@@ -618,23 +458,27 @@ TEST(CommandTest, integralOnlyTest)
...
+
+  // Set only integral gain
+  Pid pid(0.0, 1.0, 0.0, u_max, u_min, antiwindup_strat);
   double cmd = 0.0;
 
   // If initial error = 0, i-gain = 1, dt = 1
   cmd = pid.compute_command(-0.5, 1.0);
-  // Then expect command = error
+  // Then expect command = 0.0, since the integral contribution is calculated
+  // with backwards Euler (i.e. it is calculated after the command is computed)
```
christophfroehlich: isn't this wrong now, see my comment in pid.cpp?

control_toolbox/test/pid_tests.cpp
```diff
@@ -618,23 +458,27 @@ TEST(CommandTest, integralOnlyTest)
...
   EXPECT_EQ(-0.5, cmd);
 
   // If call again with same arguments
   cmd = pid.compute_command(-0.5, 1.0);
-  // Then expect the integral part to double the command
   EXPECT_EQ(-1.0, cmd);
 
   // Call again with no error
   cmd = pid.compute_command(0.0, 1.0);
-  // Expect the integral part to keep the previous command because it ensures error = 0
```
christophfroehlich: why removing the comments?

control_toolbox/test/pid_tests.cpp
```diff
@@ -643,13 +487,11 @@ TEST(CommandTest, integralOnlyTest)
 
   // Finally call again with positive error to see if the command changes in the opposite direction
   cmd = pid.compute_command(1.0, 1.0);
-  // Expect that the command is cleared since error = -1 * previous command, i-gain = 1, dt = 1
   EXPECT_EQ(0.0, cmd);
 
   // If initial error = 0, i-gain = 1, dt = 1
-  cmd = pid.compute_command(-0.5, 1.0);
-  // Then expect command = error
-  EXPECT_EQ(-0.5, cmd);
```
christophfroehlich: why changing this?

control_toolbox/test/pid_tests.cpp
```diff
@@ -658,17 +500,15 @@ TEST(CommandTest, integralOnlyTest)
...
   // we expect the command to be 0 if update is called error = 0
   pid.reset(false);
-  cmd = pid.compute_command(0.0, 1.0);
-  EXPECT_EQ(0.0, cmd);
+  cmd = pid.compute_command(-0.5, 1.0);
+  EXPECT_EQ(-0.5, cmd);
 
   // If initial error = 0, i-gain = 1, dt = 1
-  cmd = pid.compute_command(-0.5, 1.0);
-  // Then expect command = error
```
christophfroehlich: why changing this? If so, the comment in L505 is wrong.

control_toolbox/test/pid_tests.cpp
christophfroehlich, line 1: do we need the `#pragma` on top of the file?

christophfroehlich (changes requested):
Thanks a lot, I made some comments in the code.
Please also remove the default constructor of the Gains struct, I think it's a good time now.

Could you please also open follow-up PRs for the ros2_controllers? (see failing downstream tests).

control_toolbox/src/pid.cpp
ViktorCVS: Without anti‑windup, the integral term can be calculated before the first cmd (since it's not dependent on cmd_ ‑ cmd_unsat_); I believe it's more aligned with theory.

control_toolbox/test/pid_tests.cpp
ViktorCVS: Yes. Fixed in https://github.com/ros-controls/control_toolbox/pull/436/commits/b16628e745210208d6a47826b29d51182a35293e

control_toolbox/test/pid_tests.cpp
ViktorCVS: It wasn't intentional. Fixed in https://github.com/ros-controls/control_toolbox/pull/436/commits/b16628e745210208d6a47826b29d51182a35293e

control_toolbox/test/pid_tests.cpp
ViktorCVS: It wasn't intentional. Fixed in commit https://github.com/ros-controls/control_toolbox/pull/436/commits/b16628e745210208d6a47826b29d51182a35293e

control_toolbox/test/pid_tests.cpp
ViktorCVS: It wasn't intentional. Fixed in commit https://github.com/ros-controls/control_toolbox/commit/b16628e745210208d6a47826b29d51182a35293e

control_toolbox/test/pid_tests.cpp
ViktorCVS: Not anymore, fixed in https://github.com/ros-controls/control_toolbox/pull/436/commits/2b8dd591c49806fc5df9b816d3770cbc0b0f208f

control_toolbox/src/pid.cpp
ViktorCVS: And I guess I'll do something similar with the anti‑windup strategies. Although the cmd_ ‑ cmd_unsat_ calculation has to be done after cmd is calculated, we don't need to wait for it to calculate the partial i_term, making it more aligned with the block diagram. I'll open a new PR after this cleanup to better discuss this change.

christophfroehlich: I'm not sure if this has a notable effect in control loops, but the difference of the integral action implementation depending on the antiwindup strategy sounds strange (and makes writing tests difficult: one strategy has a direct feedthrough, another hasn't). Maybe you leave it where it is currently, and open a new PR to discuss this as you proposed?

ViktorCVS: Just to make sure, should I put the NONE strategy back together with the other strategies?

christophfroehlich: Lets ask [matlab help](https://de.mathworks.com/help/control/ref/pid.html) as a source: Forward euler is the default, but both is valid.
If you change it to backward euler, it will again break the tests I had to fix with https://github.com/ros-controls/ros2_controllers/pull/1824

We could make a new parameter to switch between both behaviors, if we can't decide on the better one.

ViktorCVS: I agree with “forward.” Although both are valid, “forward” is more natural. I'll leave it this way in this PR since the tests are ok right now, and in the next PR I'll change the others.

christophfroehlich: but your proposal now is backward?
Forward Euler:
x[n+1]=x[n]+T⋅u[n]
Backward Euler:
x[n+1]=x[n]+T⋅u[n+1]

ViktorCVS: Yes, since the purpose of this PR is to remove the deprecated parameters, I can leave it as ‘backward’ for now and change it in the next one ( strategy::none keeps with ‘forward’ to maintain behaviour).

saikishor: > Lets ask [matlab help](https://de.mathworks.com/help/control/ref/pid.html) as a source: Forward euler is the default, but both is valid. If you change it to backward euler, it will again break the tests I had to fix with [ros-controls/ros2_controllers#1824](https://github.com/ros-controls/ros2_controllers/pull/1824)
> 
> We could make a new parameter to switch between both behaviors, if we can't decide on the better one.

I agree with @christophfroehlich here, let's maintain the ForwardEuler as default one as it is coherent with many other PID implementations. 

I also like the idea of having a parameter, so we can be able to choose between the behaviours.

bijoua29:
I had a couple of comments:

1. This is a result of the previous PR but the update of the i-term is delayed by one sample iteration since it is updated after the command is calculated. It may not be a big deal in the big picture, but it seems like if the error changes the changes should be applied in the same sample iteration not the next sample. It seems like the code could be modified to apply that change immediately although it might make the code slightly more complex. It just leaves a slightly bad taste when the changes are not applied in the current iteration.

2. I am probably in the minority but I think there are use cases where you want to limit the contribution of the i-term to the control command i.e. there are cases where you might still want a i_limit (min/max) outside of the anti-windup strategy.

christophfroehlich:
@ViktorCVS did you close this on purpose? I think it was almost ready for merge?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -520,7 +391,7 @@ class Pid
         tracking_time_constant parameter to tune the anti-windup behavior.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
+   * \note New gains are not applied if u_min > u_max.
```
christophfroehlich: what about i_min > i_max?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -639,7 +461,7 @@ class Pid
         tracking_time_constant parameter to tune the anti-windup behavior.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if i_min_ > i_max_ or u_min > u_max
+   * \note New gains are not applied if u_min > u_max
```
christophfroehlich: what about i_min > i_max?

control_toolbox/src/pid.cpp
```diff
@@ -409,12 +306,16 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
     {
       i_term_ += dt_s * (gains_.i_gain_ * error +
                          1 / gains_.antiwindup_strat_.tracking_time_constant * (cmd_ - cmd_unsat_));
+
+      i_term_ = std::clamp(i_term_, gains_.i_min_, gains_.i_max_);
```
christophfroehlich:
```suggestion
```

control_toolbox/src/pid.cpp
```diff
@@ -409,12 +306,16 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
     }
     else if (gains_.antiwindup_strat_.type == AntiWindupStrategy::CONDITIONAL_INTEGRATION)
     {
       if (!(!is_zero(cmd_unsat_ - cmd_) && error * cmd_unsat_ > 0))
       {
         i_term_ += dt_s * gains_.i_gain_ * error;
+
+        i_term_ = std::clamp(i_term_, gains_.i_min_, gains_.i_max_);
```
christophfroehlich:
```suggestion
```

christophfroehlich: Can't we move that outside of the if/else construct? It is valid for all strategies, right?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -135,23 +121,15 @@ struct AntiWindupStrategy
...
         "(tracking_time_constant)");
     }
-    if (type == LEGACY && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
+    if (type != NONE && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
     {
       throw std::invalid_argument(
         fmt::format(
-          "AntiWindupStrategy 'legacy' requires i_min < i_max and to be finite (i_min: {}, i_max: "
-          "{})",
+          "AntiWindupStrategy requires i_min < i_max and to be finite (i_min: {}, i_max: {})",
```
```suggestion
          "PID requires i_min < i_max if limits are finite (i_min: {}, i_max: {})",
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -135,23 +121,15 @@ struct AntiWindupStrategy
         "AntiWindupStrategy 'back_calculation' requires a valid positive tracking time constant "
         "(tracking_time_constant)");
     }
-    if (type == LEGACY && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
+    if (type != NONE && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
```
```suggestion
    if ((i_min > i_max) && ( !std::isfinite(i_min) || !std::isfinite(i_max)))
```
i_min and i_max is valid for all integration/antiwindup techniques?

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -624,11 +588,11 @@ TEST(PidParametersTest, GetParametersFromParams)
 
   rclcpp::Parameter param_i_clamp_max;
   ASSERT_TRUE(node->get_parameter("i_clamp_max", param_i_clamp_max));
-  EXPECT_TRUE(std::isnan(param_i_clamp_max.get_value<double>()));
+  ASSERT_TRUE(std::isinf(param_i_clamp_max.get_value<double>()));
```
christophfroehlich:
```suggestion
  EXPECT_TRUE(std::isinf(param_i_clamp_max.get_value<double>()));
```

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -624,11 +588,11 @@ TEST(PidParametersTest, GetParametersFromParams)
...
 
   rclcpp::Parameter param_i_clamp_min;
   ASSERT_TRUE(node->get_parameter("i_clamp_min", param_i_clamp_min));
-  EXPECT_TRUE(std::isnan(param_i_clamp_min.get_value<double>()));
+  ASSERT_TRUE(std::isinf(param_i_clamp_min.get_value<double>()));
```
christophfroehlich:
```suggestion
  EXPECT_TRUE(std::isinf(param_i_clamp_min.get_value<double>()));
```

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -608,7 +572,7 @@ TEST(PidParametersTest, GetParametersFromParams)
   const bool ACTIVATE_STATE_PUBLISHER = false;
   TestablePidROS pid(node, "", "", ACTIVATE_STATE_PUBLISHER);
 
-  ASSERT_FALSE(pid.initialize_from_ros_parameters());
```
christophfroehlich, line 611: why was this false before?

christophfroehlich (changes requested):
Thanks a lot for the follow-up. One question: Shouldn't we allow i_min/i_max also for the NONE strategy?

control_toolbox/src/pid.cpp
```diff
@@ -354,35 +276,11 @@ double Pid::compute_command(double error, double error_dot, const double & dt_s)
...
-    gains_.antiwindup_strat_.type == AntiWindupStrategy::LEGACY)
-  {
-    // Limit i_term so that the limit is meaningful in the output
-    cmd_unsat_ = p_term + std::clamp(i_term_, gains_.i_min_, gains_.i_max_) + d_term;
-  }
-  else
-  {
-    cmd_unsat_ = p_term + i_term_ + d_term;
-  }
+  cmd_unsat_ = p_term + i_term_ + d_term;
```
bijoua29, line 283: Weren't we going to consider doing forward integration for integral term instead of backward integration?

christophfroehlich: we have it consistent for all antiwindup strategies now, I think the selectable integration method will be handled in a different PR.

christophfroehlich:
> However, I trust your judgment about what's best for this package. Should I proceed with adding it?

Yes please do so.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: This comment refers to the initialize() method, which doesn't have i_max or i_min parameters.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: This comment refers to the set_gains() method, which doesn't have i_max or i_min parameters.

control_toolbox/src/pid.cpp
ViktorCVS: Is this comment the same as the next one?

ViktorCVS: Thanks! Solved in commit 2a77d33

control_toolbox/src/pid.cpp
ViktorCVS: Thanks! Solved in commit 2a77d33

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: isn't i_min/i_max part of AntiWindupStrategy?

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: isn't i_min/i_max part of AntiWindupStrategy?

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Thanks! Solved in commit f278c30

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Thanks! Solved in commit f278c30

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Yes! I didn't know that it would be required in this case. Fixed in commit 949098a

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Yes! I didn't know that it would be required in this case. Fixed in commit 949098a

control_toolbox/test/pid_ros_parameters_tests.cpp
ViktorCVS: I believe it was false because i_max and i_min began as NaN, so it failed before when those values weren’t set. Now, with them starting as infinity, this method returns true.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: This one solve it 7e3cee2

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -121,12 +121,11 @@ struct AntiWindupStrategy
         "AntiWindupStrategy 'back_calculation' requires a valid positive tracking time constant "
         "(tracking_time_constant)");
     }
-    if (type != NONE && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
+    if (std::isfinite(i_min) && std::isfinite(i_max) && !(i_min < i_max))
```
```suggestion
    if (i_min >= i_max)
```
what happens if anyone sets the parameter i_min = +inf and i_max=0 (any finite number)? i_max=-inf and i_min=-inf is undefined behavior as well? http://en.cppreference.com/w/cpp/algorithm/clamp.html

do we have to check for finiteness at all? I think the <> operators are well defined for infinity, but not for NaN (maybe we have to check for NaN, but we would need that for every param?)

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -242,7 +242,7 @@ class PidROS
         tracking_time_constant parameter to tune the anti-windup behavior.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if u_min > u_max.
+   * \note New gains are not applied if i_min_ > i_max_ or u_min_ > u_max_.
```
christophfroehlich:
```suggestion
   * \note New gains are not applied if antiwindup_strat.i_min > antiwindup_strat.i_max or u_min > u_max.
```

control_toolbox/include/control_toolbox/pid_ros.hpp
```diff
@@ -169,7 +169,7 @@ class PidROS
    * \param save_i_term save integrator output between resets.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if u_min_ > u_max_.
+   * \note New gains are not applied if i_min_ > i_max_ or u_min_ > u_max_.
```
christophfroehlich:
```suggestion
   * \note New gains are not applied if antiwindup_strat.i_min > antiwindup_strat.i_max or u_min > u_max.
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -391,7 +390,7 @@ class Pid
         tracking_time_constant parameter to tune the anti-windup behavior.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if u_min > u_max.
+   * \note New gains are not applied if i_min_ > i_max_ or u_min_ > u_max_.
```
```suggestion
   * \note New gains are not applied if antiwindup_strat.i_min > antiwindup_strat.i_max or u_min > u_max.
```

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -461,7 +460,7 @@ class Pid
         tracking_time_constant parameter to tune the anti-windup behavior.
    * \return True if all parameters are successfully set, False otherwise.
    *
-   * \note New gains are not applied if u_min > u_max
+   * \note New gains are not applied if i_min_ > i_max_ or u_min_ > u_max_.
```
```suggestion
   * \note New gains are not applied if antiwindup_strat.i_min > antiwindup_strat.i_max or u_min > u_max.
```

christophfroehlich (changes requested):
One last comment about the input checks, and sorry for being pedantic about the docstrings ;)

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: @ViktorCVS what do you think about this?

ViktorCVS: I do agree with you! And I don't believe that a NaN would be a problem here or with other params. I'll commit this sugestion.

christophfroehlich (approved):
Thanks, LGTM!

PR #443, opened 2025-08-06
Declare missing parameters for PID

### LINKED ISSUE #442: Latest Kilted sync generates exception from PID controller
**Describe the bug**
I just recently updated to latest Kilted sync from 2025-07-28. I am running PidController from ros2 controllers which uses PidROS from control_toolbox, version 5.6.0. When I enable PID control, I believe I am setting some gains. However, I am getting the following exception when I do that:

[ros2_control_node-2] [ERROR] [1754410873.174428177] [controller_manager]: Caught exception of type : N6rclcpp10exceptions29ParameterNotDeclaredExceptionE while updating controller 'my_controller': parameter 'gains.my_controller/xxx.saturation' cannot be set because it was not declared

**To Reproduce**
1. Activate a PidController controller
2. Change pid gains on the controller
5. See above exception

**Expected behavior**

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Version: Kilted sync from 2025-07-28

**Additional context**
It seems that the PidROS::initialize_from_ros_parameters() method is not declaring the 'saturation' parameter which may be causing this exception. I also noticed that it is not declaring the 'activate_state_publisher' parameter which may be a bug too.

christophfroehlich:
Resolves #442 

I fixed the tests, we haven't tested all the parameters with `initialize_from_ros_parameters`

### DIFF
--- a/control_toolbox/src/pid_ros.cpp
+++ b/control_toolbox/src/pid_ros.cpp
@@ -394,7 +394,9 @@ bool PidROS::initialize_from_args(
         rclcpp::ParameterValue(antiwindup_strat.tracking_time_constant));
       declare_param(
         param_prefix_ + "error_deadband", rclcpp::ParameterValue(antiwindup_strat.error_deadband));
-      declare_param(param_prefix_ + "saturation", rclcpp::ParameterValue(true));
+      declare_param(
+        param_prefix_ + "saturation",
+        rclcpp::ParameterValue(std::isfinite(gains.u_max_) || std::isfinite(gains.u_min_)));
       declare_param(
         param_prefix_ + "antiwindup_strategy",
         rclcpp::ParameterValue(gains.antiwindup_strat_.to_string()));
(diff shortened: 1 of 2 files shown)


control_toolbox/src/pid_ros.cpp
```diff
@@ -260,6 +260,7 @@ bool PidROS::initialize_from_ros_parameters()
   u_max = UMAX_INFINITY;
   u_min = -UMAX_INFINITY;
   bool antiwindup = false;
+  bool saturation = false;
```
saikishor:
```suggestion
  bool saturation = true;
```

Shouldn't the default be true?

https://github.com/ros-controls/control_toolbox/blob/2ad92f1f225918c9c192b8a7e9ebfa3b959ddd10/control_toolbox/src/pid_ros.cpp#L388

TO be coherent with the above one

christophfroehlich: I'd change it in initialize_from_args to be true only if u_min or u_max argument is finite.
for ROS parameters, idk if it should be default false or default true? or use the same logic, depending on u_clamp_max/min parameter?

saikishor (dismissed):

Issue #469, opened 2025-09-11
Add option to PID for Forward/Backward integration


christophfroehlich:
See https://github.com/ros-controls/control_toolbox/pull/436#discussion_r2244745868 for reference

Theonewhomadethings:
Hi @christophfroehlich, after PR (https://github.com/ros-controls/control_toolbox/pull/484) and issue (https://github.com/ros-controls/control_toolbox/issues/409) are closed could I tackle this as a second issue?

christophfroehlich:
@ViktorCVS is already working on it with #471

PR #470, opened 2025-09-14
Feature/469 forward calc

silanus23:
@christophfroehlich 
This issue is about #469 . I have created tests too. Directly took the  math from https://de.mathworks.com/help/control/ref/pid.html . Previous error passing system could have been better but I don't want to tweak system too much.


christophfroehlich (changes requested):
I'm not sure what you try to implement here. My issue was about adding options for the discretization methods of the integral and differential part, see `BackwardEuler` , `ForwardEuler`  of IFormula and DFormula in the matlab help.

silanus23:
@christophfroehlich
You're right, it looks like I misunderstood this issue. Sorry.  I also see that this has now been implemented correctly in other PR before me. 
Thanks for your time and feedback.

PR #471, opened 2025-09-16
Add derivative filter and discretization

### LINKED ISSUE #469: Add option to PID for Forward/Backward integration

ViktorCVS:
### **Overview**

Adds a configurable derivative filter. It also introduces separate discretization method selectors for the I and D terms. Forward Euler remains the default to preserve existing behavior.

* Added a first-order derivative filter (`tf`) applied to the D term.
* Introduced per-term discretization selectors: `i_method` and `d_method` (e.g., `"forward_euler"`, `"backward_euler"`, `"trapezoidal"`), with `"forward_euler"` as the default.
* Implemented discrete D-term updates for the supported methods, including filtered forms; falls back to unfiltered behavior when `tf == 0`.
* Added internal state (`d_term_last_`, `d_error_last_`, `i_term_last_`, `aw_term_last_`) to support trapezoidal (Tustin) and filtered updates.

### **About tests**

The packages compile correctly and have passed the *pre‑commit* and *colcon* tests. New tests are being added based on formula values to compare with those from `compute_command`.

### **Related PR's**

- https://github.com/ros-controls/ros2_controllers/pull/1909

### **Attachments**

Full derivation of the discrete-time PID used in this PR. The document shows the step-by-step development of the equations.

[ros2_control_new_discretization_feature.pdf](https://github.com/user-attachments/files/25522067/ros2_control_new_discretization_feature.pdf)

### Simulation

I developed a small simulation that lets you compare a controlled "continuous-time" plant with its discretized counterpart using the changes in this PR. The source file is available here:

https://github.com/ViktorCVS/ros2_control_sim
or
[sim_ros2_control.cpp](https://github.com/user-attachments/files/25722035/sim_ros2_control.cpp)

#### How to use

1. Place this file in `control_toolbox/control_toolbox/src`.

2. Update `CMakeLists.txt` (around line 163) to add the target:

   ```cmake
   add_executable(sim src/sim_ros2_control.cpp)
   target_link_libraries(sim control_toolbox)
   install(TARGETS sim
     DESTINATION lib/${PROJECT_NAME}
   )
   ```

   It should look like this:

   image

3. Build and run the simulation:

   ```bash
   colcon build --packages-select control_toolbox
   ros2 run control_toolbox sim
   ```

**Note:** This simulation uses Gnuplot. Please ensure it is installed before proceeding.
**Note:** Check the simulation parameters. As currently configured, each array holds about 10 million values (~80 MB), for a total of ~480 MB.

### **Final notes**


ViktorCVS:
Hi, @christophfroehlich! I still need to fix the existing tests and add new ones. However, the equations are quite large, and the compute_command function has grown a lot. Do you think it’s acceptable (with better organization), or should something be changed?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -329,21 +377,25 @@ class Pid
...
     double d_gain_ = 0.0; /**< Derivative gain. */
+    double tf_ = 0.0;     /**< Derivative filter time constant. */
     double i_max_ =
       std::numeric_limits<double>::infinity(); /**< Maximum allowable integral term. */
     double i_min_ =
       -std::numeric_limits<double>::infinity(); /**< Minimum allowable integral term. */
     double u_max_ = std::numeric_limits<double>::infinity();  /**< Maximum allowable output. */
     double u_min_ = -std::numeric_limits<double>::infinity(); /**< Minimum allowable output. */
     AntiWindupStrategy antiwindup_strat_;                     /**< Anti-windup strategy. */
+    std::string i_method_ = "forward_euler";                  /**< Integration method. */
```
christophfroehlich: why have you chosen a std::string here? Wouldn't a simple enum be more lightweight here? Only thing we loose will be the human readable value in the print method without adding additional logic there. alternative would be a struct like the andiwindup strategy.

control_toolbox/src/pid_ros.cpp
```diff
@@ -594,6 +625,11 @@ void PidROS::set_parameter_event_callback()
           gains.d_gain_ = parameter.get_value<double>();
           changed = true;
         }
+        else if (param_name == param_prefix_ + "derivative_filter_time")
```
christophfroehlich, line 556: I suppose you have not added the integration/derivative method here on purpose?

ViktorCVS: Yes, for now I don’t think anyone would modify the discretization methods during runtime.

control_toolbox/include/control_toolbox/pid.hpp
ViktorCVS: Thanks! I change it to struct in commit 2b7a909

ViktorCVS:
@christophfroehlich  @saikishor  Hi! could you please review this PR?

PR #481, opened 2025-09-18
Fix source build workflow on rolling

christophfroehlich:
I missed something..


christophfroehlich:
@mergifyio backport mergify/bp/humble/pr-478 mergify/bp/jazzy/pr-478

PR #484, opened 2025-09-21
Increase PID ROS wrapper test coverage 

Theonewhomadethings:
# Overview

This PR increases test coverage for control_toolbox::PidROS, focusing on methods that previously had little or no coverage. (as described in https://github.com/ros-controls/control_toolbox/issues/409)

- void PidROS::reset()
- void PidROS::reset(bool save_i_term)
- double PidROS::compute_command(double error, double error_dot, const rclcpp::Duration & dt)
- void PidROS::set_current_cmd(double cmd) { pid_.set_current_cmd(cmd); }
- double PidROS::get_current_cmd() { return pid_.get_current_cmd(); }
- void PidROS::print_values()

# What changed

files changed are:

- control_toolbox/control_toolbox/test/pid_ros_publisher_tests.cpp
- control_toolbox/control_toolbox/test/pid_ros_parameters_tests.cpp

Added 6 tests to cover these methods in the below files

In control_toolbox/test/pid_ros_parameters_tests.cpp, we added five tests: two I-only tests that check reset() and reset(bool) either clear or keep the integral term; one PI test that calls print_values() and captures the log output; and two P-only tests that cover get_current_cmd() and set_current_cmd(), using tight limits so the results are predictable.

In control_toolbox/test/pid_ros_publisher_tests.cpp, we added one test that runs the publisher end-to-end on a LifecycleNode and confirms a PidState message is published.

Before every commit I made sure to build the package locally, run colcon test and do pre-commit commands. All tests passed and no formatting was flagged.

# Notes for Reviewer

This is my first contribution to ros2_control, so I am very open to feedback and am happy to make any changes that are requested. 

I can remove comments after a initial review has been made?


control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -615,6 +615,259 @@ TEST(PidParametersTest, GetParametersFromParams)
...
+  const auto dt = rclcpp::Duration::from_seconds(1.0);
+
+  // Prime and snapshot
+  (void)pid.compute_command(1.0, dt);
+  (void)pid.compute_command(1.0, dt);
+  const double I_snapshot = pid.compute_command(0.0, dt);
+  EXPECT_GT(I_snapshot, 0.0);
+
+  // reset(false) clears integral
+  pid.reset(/*save_i_term=*/false);
```
saikishor:
```suggestion
  pid.reset(false);
```

Do we need to add the comment part? It is already evident from the API

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -615,6 +615,259 @@ TEST(PidParametersTest, GetParametersFromParams)
...
+  EXPECT_DOUBLE_EQ(0.0, after_clear);
+
+  // Re-prime and snapshot again
+  (void)pid.compute_command(1.0, dt);
+  (void)pid.compute_command(1.0, dt);
+  const double I_snapshot2 = pid.compute_command(0.0, dt);
+  EXPECT_GT(I_snapshot2, 0.0);
+
+  // reset(true) retains integral
+  pid.reset(/*save_i_term=*/true);
```
saikishor:
```suggestion
  pid.reset(true);
```

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -615,6 +615,259 @@ TEST(PidParametersTest, GetParametersFromParams)
...
+  const double after_retain = pid.compute_command(0.0, dt);
+  EXPECT_DOUBLE_EQ(I_snapshot2, after_retain);
+}
+
+TEST(PidParametersTest, PrintValuesLogsExpectedContent)
```
saikishor, line 701: Wow!
I would have never thought of handling it this way.

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -615,6 +615,259 @@ TEST(PidParametersTest, GetParametersFromParams)
...
+
+  // Simple, deterministic P-only controller with tight output limits to exercise saturation.
+  control_toolbox::AntiWindupStrategy anti;
+  anti.type = control_toolbox::AntiWindupStrategy::NONE;
+  const double U_MAX = 5.0, U_MIN = -5.0;
+
+  TestablePidROS pid(
+    node, /*param_prefix=*/"", /*topic_prefix=*/"", /*activate_state_publisher=*/false);
+  ASSERT_TRUE(pid.initialize_from_args(/*p=*/2.0, /*i=*/0.0, /*d=*/0.0, U_MAX, U_MIN, anti,
+                                       /*save_i_term=*/false));
```
saikishor: I don't know if it makes sense to mention each and every argument as comments here

control_toolbox/test/pid_ros_parameters_tests.cpp
```diff
@@ -615,6 +615,259 @@ TEST(PidParametersTest, GetParametersFromParams)
...
+{
+  auto node = std::make_shared<rclcpp::Node>("pidros_set_current_cmd_basic_test");
+
+  // Simple P-only controller with tight limits so compute() is always finite.
+  control_toolbox::AntiWindupStrategy anti;
+  anti.type = control_toolbox::AntiWindupStrategy::NONE;
+  const double U_MAX = 5.0, U_MIN = -5.0;
+
+  TestablePidROS pid(node, "", "", /*activate_state_publisher=*/false);
+  ASSERT_TRUE(pid.initialize_from_args(/*p=*/2.0, /*i=*/0.0, /*d=*/0.0, U_MAX, U_MIN, anti,
```
saikishor: Same here

control_toolbox/test/pid_ros_parameters_tests.cpp
Theonewhomadethings: Thanks for the suggestion! You’re right—the inline named-argument comments aren’t needed here since the API is already clear. I’ve updated the calls to use plain booleans for readability. Addressed in commit https://github.com/ros-controls/control_toolbox/pull/484/commits/fcaba1971d6fb90887f9bf10b95a6f2ca45b2760

control_toolbox/test/pid_ros_parameters_tests.cpp
Theonewhomadethings: Addressed in commit https://github.com/ros-controls/control_toolbox/commit/fcaba1971d6fb90887f9bf10b95a6f2ca45b2760

control_toolbox/test/pid_ros_parameters_tests.cpp
Theonewhomadethings: Thanks! For PrintValuesLogsExpectedContent I wanted a clean, deterministic way to assert on print_values() without scraping stdout, so I temporarily hook the rcutils output handler, filter by our logger name, and then restore the previous handler at the end. This keeps the test self-contained and prevents side effects on other tests.

control_toolbox/test/pid_ros_parameters_tests.cpp
Theonewhomadethings: Good point—agree those inline /*p=*/-style comments are unnecessary. I’ve removed them for clarity.  Addressed in commit https://github.com/ros-controls/control_toolbox/pull/484/commits/0c6fd48ca93e99a71c79347e4cac0132ffc066a6

control_toolbox/test/pid_ros_parameters_tests.cpp
Theonewhomadethings: Addressed in commit https://github.com/ros-controls/control_toolbox/pull/484/commits/9a05ae8b214bc14843797dd808282ba3abaaffc6

control_toolbox/test/pid_ros_publisher_tests.cpp
```diff
@@ -214,6 +214,58 @@ TEST(PidPublisherTest, PublishTest_local_prefix)
...
+  const rclcpp::Duration dt(0, 100000000);  // 0.1 s
+
+  // Expected: 1*e + 0*i + 1*e_dot = 5
+  const double cmd = pid_ros.compute_command(error, error_dot, dt);
+  EXPECT_EQ(5.0, cmd);
+
+  // Wait for the message to be delivered
+  for (size_t i = 0; i < ATTEMPTS && !callback_called; ++i)
+  {
+    rclcpp::spin_some(node);
```
christophfroehlich: this is deprecated, see https://github.com/ros-controls/control_toolbox/issues/488

control_toolbox/test/pid_ros_publisher_tests.cpp
```diff
@@ -214,6 +214,58 @@ TEST(PidPublisherTest, PublishTest_local_prefix)
...
+    std::this_thread::sleep_for(DELAY);
+  }
+
+  ASSERT_TRUE(callback_called);
+  ASSERT_NE(nullptr, last_state_msg);
+
+  // Basic content checks (PidState fields)
```
christophfroehlich, line 292: shouldn't we add this to the other tests, too?

christophfroehlich (commented):
Overall LGTM, some minor comments

control_toolbox/test/pid_ros_publisher_tests.cpp
Theonewhomadethings: thanks for spotting this, fixed in https://github.com/ros-controls/control_toolbox/pull/484/commits/b3b4d5dd35022a6a4759180eb25194aa082242b3

control_toolbox/test/pid_ros_publisher_tests.cpp
Theonewhomadethings: Yep we can, added the asserts in the other tests in pid_ros_publisher_tests.cpp via https://github.com/ros-controls/control_toolbox/pull/484/commits/83079f7bc9f3b0d5a455b5d9b25a7631521a2030, was this what you had in mind?

control_toolbox/test/pid_ros_publisher_tests.cpp
Theonewhomadethings: @christophfroehlich  Out of curiosity do you want me to patch all the tests in pid_ros_publisher_tests.cpp and pid_ros_parameters_tests.cpp to replace the deprecated approach in this PR, or do you want to leave issue (https://github.com/ros-controls/control_toolbox/issues/488) open for a different first time contributor to do this?

christophfroehlich: you can leave it for a first-time contributor. I just mentioned it to avoid adding another one with this PR

control_toolbox/test/pid_ros_publisher_tests.cpp
christophfroehlich: yes, thx!

PR #499, opened 2025-10-03
Fix ambiguous constructor overload

christophfroehlich:
On jazzy and before the cleanup #494, we had an ambiguous overload. I suggest changing the same on kilted+rolling to have the same API everywhere.

docstring was wrong anyways :see_no_evil: 

```
/workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/test/pid_ros_parameters_tests.cpp: In instantiation of ‘TestablePidROS::TestablePidROS(std::shared_ptr<_Tp>, std::string) [with NodeT = rclcpp::Node; std::string = std::__cxx11::basic_string<char>]’:
/workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/test/pid_ros_parameters_tests.cpp:136:30:   required from here
/workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/test/pid_ros_parameters_tests.cpp:50:51: error: call of overloaded ‘PidROS(std::shared_ptr<rclcpp::Node>&, std::string&)’ is ambiguous
   50 |   : control_toolbox::PidROS(node_ptr, param_prefix)
      |                                                   ^
In file included from /workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/test/pid_ros_parameters_tests.cpp:16:
/workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/include/control_toolbox/pid_ros.hpp:96:12: note: candidate: ‘control_toolbox::PidROS::PidROS(std::shared_ptr<_Tp>, const std::string&, const std::string&) [with NodeT = rclcpp::Node; std::string = std::__cxx11::basic_string<char>]’
   96 |   explicit PidROS(
      |            ^~~~~~
/workspaces/ros2_jazzy_ws/src/control_toolbox/control_toolbox/include/control_toolbox/pid_ros.hpp:71:89: note: candidate: ‘control_toolbox::PidROS::PidROS(std::shared_ptr<_Tp>, std::string, bool) [with NodeT = rclcpp::Node; std::string = std::__cxx11::basic_string<char>]’
   71 |   [[deprecated("Use overloads with explicit prefixes for params and topics")]] explicit PidROS(
      |                                                                                         ^~~~~~
```

### DIFF
--- a/control_toolbox/include/control_toolbox/pid_ros.hpp
+++ b/control_toolbox/include/control_toolbox/pid_ros.hpp
@@ -59,31 +59,13 @@ public:
    * The node is passed to this class to handler the ROS parameters, this class allows
    * to add a prefix to the pid parameters
    *
-   * \param node ROS node
-   * \param prefix prefix to add to the pid parameters.
-   *               Per default is prefix interpreted as prefix for topics.
-   * \param prefix_is_for_params provided prefix should be interpreted as prefix for parameters.
-   *        If the parameter is `true` then "/" in the middle of the string will not be replaced
-   *        with "." for parameters prefix. "/" or "~/" at the beginning will be removed.
-   *
-   */
-  template <class NodeT>
-  explicit PidROS(std::shared_ptr<NodeT> node_ptr, const std::string & param_prefix)
-  : PidROS(
-      node_ptr->get_node_base_interface(), node_ptr->get_node_logging_interface(),
-      node_ptr->get_node_parameters_interface(), node_ptr->get_node_topics_interface(),
-      param_prefix, "", false)
-  {
-  }
-  /*!
-   * \brief Constructor of PidROS class.
-   *
-   * The node is passed to this class to handler the ROS parameters, this class allows
-   * to add a prefix to the pid parameters
-   *
    * \param node Any ROS node
    * \param param_prefix prefix to add to the pid parameters.
-   * \param topic_prefix prefix to add to the state publisher. If it starts with `~/`, topic will be local under the namespace of the node. If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   * \param topic_prefix prefix to add to the state publisher.
+   *      If it starts with `~/`, topic will be local under the namespace of the node.
+   *      If it starts with `/` or an alphanumeric character, topic will be in global namespace.
+   *
+   * state publisher is not activated if topic_prefix is empty,
    *
    */
   template <class NodeT>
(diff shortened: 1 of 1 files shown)


saikishor (approved):
Looks good to me, but I'm afraid that it might break many packages. Should we deprecate it before?

Let's deprecate this here and the other one in Jazzy? Or let's remove as you proposed

christophfroehlich:
It was not working before the cleanup #494 anyways, right? The working version was just released a few days ago; and there is no use of this class on the ros buildfarm afaik. I don't see a high risk here

saikishor:
> It was not working before the cleanup #494 anyways, right? The working version was just released a few days ago; and there is no use of this class on the ros buildfarm afaik. I don't see a high risk here

**These are the ambiguous constructors that can accept 2 args in Jazzy:**
https://github.com/ros-controls/control_toolbox/blob/7b819ce45fbfab117eafb8139ac86c7c93d74099/control_toolbox/include/control_toolbox/pid_ros.hpp#L70-L87

**This is the only constructor that accepts 2 args in Rolling:**
https://github.com/ros-controls/control_toolbox/blob/6829ff1451844901e7a2fea68ea1a2b81a8606db/control_toolbox/include/control_toolbox/pid_ros.hpp#L70-L77

What I mean is, if we remove the only constructor that is using the 2 args, then if some user code has this change, it will break his code. Does it make sense?

christophfroehlich:
> > It was not working before the cleanup #494 anyways, right? The working version was just released a few days ago; and there is no use of this class on the ros buildfarm afaik. I don't see a high risk here
> 
> **These are the ambiguous constructors that can accept 2 args in Jazzy:**
> 
> https://github.com/ros-controls/control_toolbox/blob/7b819ce45fbfab117eafb8139ac86c7c93d74099/control_toolbox/include/control_toolbox/pid_ros.hpp#L70-L87

but this is not functional with the error message above. this was "accidentally" fixed with #494 (5.8.0 on rolling), released just a couple of days ago:

> **This is the only constructor that accepts 2 args in Rolling:**
> 
> https://github.com/ros-controls/control_toolbox/blob/6829ff1451844901e7a2fea68ea1a2b81a8606db/control_toolbox/include/control_toolbox/pid_ros.hpp#L70-L77
> 
> What I mean is, if we remove the only constructor that is using the 2 args, then if some user code has this change, it will break his code. Does it make sense?. If so, what I'm saying is better to add a deprecation note to the 2 args constructor and then remove current non-deprecated constructor in Jazzy

This means that after #431 (5.6.0 on rolling, 4.6.0 on jazzy) it was not possible to use the 2-args constructor. So I don't think that anyone used it (no one will have used it since the 5.8.0 release this week). That's why I don't think this hurts if we remove it now.

PR #507, opened 2025-10-09
Remove deprecation of i_min/i_max  PID parameters

christophfroehlich:
History:

- we deprecated i clamping with #410 et al, which we backported to jazzy
- then we readded it/removed the deprecation on kilted/rolling with #436

No we have it on humble, deprecated on jazzy, and have it again in kilted/rolling.

I cherry-picked the changes from #436 to have that silenced and working in jazzy.

### DIFF
--- a/control_toolbox/test/pid_ros_parameters_tests.cpp
+++ b/control_toolbox/test/pid_ros_parameters_tests.cpp
@@ -760,7 +760,8 @@ TEST(PidParametersTest, PrintValuesLogsExpectedContent)
   };
 
   // Ensure our logger emits INFO
-  rcutils_logging_set_logger_level(kLoggerName, RCUTILS_LOG_SEVERITY_INFO);
+  ASSERT_EQ(
+    rcutils_logging_set_logger_level(kLoggerName, RCUTILS_LOG_SEVERITY_INFO), RCUTILS_RET_OK);
 
   // Swap in our capture handler
   prev_handler = rcutils_logging_get_output_handler();


control_toolbox/src/pid_ros.cpp
```diff
@@ -257,8 +257,8 @@ bool PidROS::initialize_from_ros_parameters()
   double p, i, d, i_max, i_min, u_max, u_min, tracking_time_constant, error_deadband;
   p = i = d = i_max = i_min = tracking_time_constant = std::numeric_limits<double>::quiet_NaN();
   error_deadband = std::numeric_limits<double>::epsilon();
-  u_max = UMAX_INFINITY;
-  u_min = -UMAX_INFINITY;
+  u_max = i_max = UMAX_INFINITY;
+  u_min = i_min = -UMAX_INFINITY;
```
ViktorCVS: Changing UMAX_INFINITY/MIN to MAX_INFINITY/MIN would improve the documentation (also update line 43).

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -135,20 +135,11 @@ struct AntiWindupStrategy
         "AntiWindupStrategy 'back_calculation' requires a valid positive tracking time constant "
         "(tracking_time_constant)");
     }
-    if (type == LEGACY && ((i_min > i_max) || !std::isfinite(i_min) || !std::isfinite(i_max)))
+    if (i_min > i_max)
```
saikishor, line 138: @christophfroehlich Shall we also add the condition for the i_min to be negative? Or not needed?

christophfroehlich: technically, it would work (std::clamp is only undefined if i_min>i_max), but it makes no sense for the PID. I'd change that on rolling and backport to all distros then.

PR #510, opened 2025-10-13
Improve PID parameter validation

christophfroehlich:
https://github.com/ros-controls/control_toolbox/pull/507#issuecomment-3389261205

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -352,7 +352,7 @@ public:
         'conditional_integration', or 'none'. Note that the 'back_calculation' strategy use the
         tracking_time_constant parameter to tune the anti-windup behavior.
    *
-   * \throws An std::invalid_argument exception is thrown if u_min > 0 or u_max < 0.
+   * \throws An std::invalid_argument exception is thrown if u_min >= u_max.
    */
   Pid(
     double p = 0.0, double i = 0.0, double d = 0.0,


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -299,12 +303,7 @@ class Pid
 
     bool validate(std::string & error_msg) const
     {
-      if (i_min_ > i_max_)
-      {
-        error_msg = fmt::format("Gains: i_min ({}) must be less than i_max ({})", i_min_, i_max_);
-        return false;
-      }
-      else if (u_min_ >= u_max_)
+      if (u_min_ >= u_max_)
```
bijoua29: Why did you remove the check instead of replacing with the i_min/i_max check done earlier?

christophfroehlich: I pushed another change, sorry. There was a duplicate check in set_gains, and in the antiwindup_strategy's validate method. I moved now everything into the antiwindup_strategy, but leave the duplicate storage for API compatibility reasons (will be removed with #512).
I also removed the input check from the constructor of the Gains struct, to have everything in a single point (in the validate() method)

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -278,33 +289,17 @@ class Pid
     : p_gain_(p),
       i_gain_(i),
       d_gain_(d),
-      i_max_(antiwindup_strat.i_max),
-      i_min_(antiwindup_strat.i_min),
       u_max_(u_max),
       u_min_(u_min),
+      i_max_(antiwindup_strat.i_max),
```
bijoua29, line 294: Is i_max_/i_min_ variables even necessary as part of Gains as it is in anti-windup strategy?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -278,33 +289,17 @@ class Pid
...
+      i_min_(antiwindup_strat.i_min),
       antiwindup_strat_(antiwindup_strat)
     {
-      if (std::isnan(u_min) || std::isnan(u_max))
```
bijoua29, line 287: Is there a reason for removing u_min/max validation in constructor?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -358,7 +352,7 @@ class Pid
         'conditional_integration', or 'none'. Note that the 'back_calculation' strategy use the
         tracking_time_constant parameter to tune the anti-windup behavior.
    *
-   * \throws An std::invalid_argument exception is thrown if u_min > u_max.
+   * \throws An std::invalid_argument exception is thrown if u_min > 0 or u_max < 0.
```
bijoua29: Should be u_min >= u_max

control_toolbox/src/pid_ros.cpp
```diff
@@ -410,12 +410,12 @@ void PidROS::print_values()
...
       << "  D Gain:       " << gains.d_gain_ << "\n"
-      << "  I Max:        " << gains.i_max_ << "\n"
-      << "  I Min:        " << gains.i_min_ << "\n"
       << "  U_Max:                  " << gains.u_max_ << "\n"
       << "  U_Min:                  " << gains.u_min_ << "\n"
-      << "  Tracking_Time_Constant: " << gains.antiwindup_strat_.tracking_time_constant << "\n"
       << "  Antiwindup_Strategy:    " << gains.antiwindup_strat_.to_string() << "\n"
+      << "  Tracking_Time_Constant: " << gains.antiwindup_strat_.tracking_time_constant << "\n"
+      << "  I Max:        " << gains.antiwindup_strat_.i_max << "\n"
+      << "  I Min:        " << gains.antiwindup_strat_.i_min << "\n"
```
bijoua29, line 418: I wonder if you can use print() method of antiwindup strategy here instead if you change it to this format.

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: no it's not, I just kept it for API backwards compatibility. On rolling see #512

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: even we check all or none of the parameters. I find it cleaner do all the validations in the validate() method.

control_toolbox/include/control_toolbox/pid.hpp
christophfroehlich: true

control_toolbox/src/pid_ros.cpp
christophfroehlich: only if I change it to return a std::stringstream instead of streaming to std::cout. Is this worth the change?

control_toolbox/include/control_toolbox/pid.hpp
bijoua29: Fair enough

control_toolbox/src/pid_ros.cpp
bijoua29: Maybe not - up to you

PR #511, opened 2025-10-14
Fix calculation of tracking time constant

bijoua29:
Fix a couple of issues with tracking time constant used for back propagation anti-windup strategy
- fix divide-by-zero bug in calculation of tracking time constant which cause time constant to be infinite and thus fail validation check and prevent updating of gains

Note: maybe there is a better way of fixing this but this resolved my issue where I was unable to change gains using rqt when i-gain was set to zero and tracking time constant was at its default of zero.

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.hpp
+++ b/control_toolbox/include/control_toolbox/pid.hpp
@@ -45,12 +45,6 @@
 
 namespace control_toolbox
 {
-template <typename T>
-inline bool is_zero(T value, T tolerance = std::numeric_limits<T>::epsilon())
-{
-  return std::abs(value) <= tolerance;
-}
-
 /**
  * \brief Antiwindup strategy for PID controllers.
  *
@@ -113,14 +107,14 @@ public:
     }
   }
 
-  void validate(const double i_gain) const
+  void validate() const
   {
     if (type == UNDEFINED)
     {
       throw std::invalid_argument("AntiWindupStrategy is UNDEFINED. Please set a valid type");
     }
     if (
-      type == BACK_CALCULATION && !is_zero(i_gain) &&
+      type == BACK_CALCULATION &&
       (tracking_time_constant < 0.0 || !std::isfinite(tracking_time_constant)))
     {
       throw std::invalid_argument(
@@ -173,6 +167,12 @@ public:
   double error_deadband =
     std::numeric_limits<double>::epsilon(); /**< Error deadband to avoid integration. */
 };
+
+template <typename T>
+inline bool is_zero(T value, T tolerance = std::numeric_limits<T>::epsilon())
+{
+  return std::abs(value) <= tolerance;
+}
 
 /***************************************************/
 /*!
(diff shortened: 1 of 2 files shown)


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -107,14 +113,14 @@ struct AntiWindupStrategy
...
-  void validate() const
+  void validate(const double i_gain) const
   {
     if (type == UNDEFINED)
     {
       throw std::invalid_argument("AntiWindupStrategy is UNDEFINED. Please set a valid type");
     }
     if (
-      type == BACK_CALCULATION &&
+      type == BACK_CALCULATION && !is_zero(i_gain) &&
```
christophfroehlich: why do we need this? if tracking_time_constant is zero and you want to set i_gain, this still works?

bijoua29: @christophfroehlich Sorry, I should have been more clear. When tracking time constant is zero, a tracking time constant is calculated. However, this calculation is where the divide-by-zero is for i-gain equal to zero and the resulting time constant is infinite resulting in this check failing and then the updated gains are rejected. With the bug fix above, you are probably right - this change is probably not necessary. I will remove it.

bijoua29: Removed in https://github.com/ros-controls/control_toolbox/pull/511/commits/dc7d72272c48149a02ea1cfc3aefc1c49160a873

christophfroehlich:
> > I see the issue with division by zero, but don't see why the check in validate() is harming? default value of tracking_time_constant is zero, so this check will not be triggered?
> > Could you add a line to the tests please, ensuring that this remains working?
> 
> A test for i-gain and time constant = 0?

yes, please!

christophfroehlich:
@bijoua29 I'm not allowed to push to your PR, but this patch should be sufficient
```
$ git diff
diff --git a/control_toolbox/test/pid_tests.cpp b/control_toolbox/test/pid_tests.cpp
index 5629cd5..24705bd 100644
--- a/control_toolbox/test/pid_tests.cpp
+++ b/control_toolbox/test/pid_tests.cpp
@@ -213,6 +213,10 @@ TEST(ParameterTest, integrationBackCalculationZeroGainTest)
   double cmd = 0.0;
   double pe, ie, de;
 
+  // with I_gain = 0, tracking_time_constant should not be calculated
+  auto gains = pid.get_gains();
+  EXPECT_DOUBLE_EQ(gains.antiwindup_strat_.tracking_time_constant, 0.0);
+
   // back_calculation
 
   cmd = pid.compute_command(-1.0, 1.0);
@@ -244,6 +248,8 @@ TEST(ParameterTest, integrationBackCalculationZeroGainTest)
   pid.get_current_pid_errors(pe, ie, de);
   EXPECT_EQ(0.0, ie);
   EXPECT_EQ(0.0, cmd);
:
```

christophfroehlich (approved):
I'll merge this and add the test in a different PR.

PR #533, opened 2025-10-24
TF prefix helper added

kuralme:
Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:

1. Limited scope. Your PR should do one thing or one set of things. Avoid adding “random fixes” to PRs. Put those on separate PRs.
2. Give your PR a descriptive title. Add a short summary, if required.
3. Make sure the pipeline is green.
4. Don’t be afraid to request reviews from maintainers.
5. New code = new tests. If you are adding new functionality, always make sure to add some tests exercising the code and serving as live documentation of your original intention.

To send us a pull request, please:

- [x] Fork the repository.
- [x] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [x] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [x] Commit to your fork using clear commit messages.
- [x] Send a pull request, answering any default questions in the pull request interface.
- [x] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.

PR related to the [issue](https://github.com/ros-controls/ros2_controllers/issues/1745)

### DIFF
--- a/control_toolbox/test/pid_ros_parameters_tests.cpp
+++ b/control_toolbox/test/pid_ros_parameters_tests.cpp
@@ -240,6 +240,9 @@ TEST(PidParametersTest, SetParametersTest)
 {
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("pid_parameters_test");
 
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
+
   TestablePidROS pid(node, "", "", false);
 
   const double P = 1.0;
@@ -295,8 +298,7 @@ TEST(PidParametersTest, SetParametersTest)
     set_result = node->set_parameter(rclcpp::Parameter("activate_state_publisher", true)));
   ASSERT_TRUE(set_result.successful);
 
-  // process callbacks
-  rclcpp::spin_some(node->get_node_base_interface());
+  executor.spin_some();
 
   // check gains were set using the parameters
   control_toolbox::Pid::Gains gains = pid.get_gains();
@@ -314,6 +316,9 @@ TEST(PidParametersTest, SetParametersTest)
 TEST(PidParametersTest, SetBadParametersTest)
 {
   rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("pid_parameters_test");
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
 
   TestablePidROS pid(node, "", "", false);
 
@@ -370,7 +375,7 @@ TEST(PidParametersTest, SetBadParametersTest)
   ASSERT_TRUE(set_result.successful);
 
   // process callbacks
-  rclcpp::spin_some(node->get_node_base_interface());
+  executor.spin_some();
 
   // check gains were NOT set using the parameters but the u_max and u_min
   // were set to infinity as saturation is false
@@ -393,7 +398,7 @@ TEST(PidParametersTest, SetBadParametersTest)
   ASSERT_TRUE(set_result.successful);
 
   // process callbacks
-  rclcpp::spin_some(node->get_node_base_interface());
+  executor.spin_some();
 
   // Setting good gains doesn't help, as the saturation is still false
   gains = pid.get_gains();
@@ -412,7 +417,7 @@ TEST(PidParametersTest, SetBadParametersTest)
   ASSERT_TRUE(set_result.successful);
 
   // process callbacks
-  rclcpp::spin_some(node->get_node_base_interface());
+  executor.spin_some();
 
   // check gains were NOT set using the parameters
   control_toolbox::Pid::Gains updated_gains = pid.get_gains();
--- a/control_toolbox/test/pid_ros_publisher_tests.cpp
+++ b/control_toolbox/test/pid_ros_publisher_tests.cpp
@@ -38,6 +38,9 @@ TEST(PidPublisherTest, PublishTest)
 
   auto node = std::make_shared<rclcpp::Node>("pid_publisher_test");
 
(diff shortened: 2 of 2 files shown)


control_toolbox/include/control_toolbox/tf_utils.hpp
```diff
@@ -0,0 +1,83 @@
...
+{
+/**
+* @brief Apply a TF prefix to a given frame.
+* @param tf_prefix_enabled Whether to apply the TF prefix
+* @param prefix TF prefix
+* @param frame Frame name
+* @param node_ns Node namespace to use as prefix if prefix is empty
+* @return The prefixed frame name if prefix is not empty, otherwise the original frame name
+*/
+inline std::string applyTFPrefix(
```
christophfroehlich:
```suggestion
inline std::string apply_tf_prefix(
```
please use snake case, this is our default style

control_toolbox/test/tf_utils_tests.cpp
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2008, Willow Garage, Inc.
```
christophfroehlich:
```suggestion
// Copyright (c) 2025, ros2_control developers
```
same here

control_toolbox/include/control_toolbox/tf_utils.hpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2008, Willow Garage, Inc.
```
christophfroehlich:
```suggestion
// Copyright (c) 2025, ros2_control developers
```
or add your proper affiliation if you have one.

https://github.com/ros-controls/ros2_controllers/pull/533 introduced this part, there is no distinct copyright notice.

control_toolbox/include/control_toolbox/tf_utils.hpp
```diff
@@ -0,0 +1,83 @@
...
+/**
+* @brief Apply a TF prefix to a given frame.
+* @param tf_prefix_enabled Whether to apply the TF prefix
+* @param prefix TF prefix
+* @param frame Frame name
+* @param node_ns Node namespace to use as prefix if prefix is empty
+* @return The prefixed frame name if prefix is not empty, otherwise the original frame name
+*/
+inline std::string apply_tf_prefix(
+  bool tf_prefix_enabled, std::string prefix, const std::string & node_ns,
```
christophfroehlich:
```suggestion
  bool tf_prefix_enabled, const std::string & prefix, const std::string & node_ns,
```

christophfroehlich: And I'm unsure if it is better removing `tf_prefix_enabled` as an argument, but just use a pattern like `tf_prefix_enabled ? apply_tf_prefix(...) : frame;` on the calling site? something like `apply_tf_prefix(false)` feels a bit strange for me. what do you think?

kuralme: Yes that makes better sense to remove that. I'll do so

kuralme: I have a question about the slash normalization design. With the latest version of helper utilized in the controllers(diff, omni, mecanum), their tests succeeds. Currently, it only removes the leading and add one trailing slash to tf prefix. But what about cases when user put multiple slashes or leading slash in the namespace?

I got the test cases for these, is this expected behaviour?
```c++
EXPECT_EQ(control_toolbox::apply_tf_prefix("robot2//", "/ns", "odom"), "robot2//odom");
EXPECT_EQ(control_toolbox::apply_tf_prefix("robot", "/ns", "/odom"), "robot//odom");
```
I can also make it ensure there is only one in between:
```c++
EXPECT_EQ(control_toolbox::apply_tf_prefix("robot2//", "/ns", "odom"), "robot2/odom");
EXPECT_EQ(control_toolbox::apply_tf_prefix("robot", "/ns", "/odom"), "robot/odom");
EXPECT_EQ(control_toolbox::apply_tf_prefix("robot/", "/ns", "/odom"), "robot/odom");
```
So, shall i change the design or leave as it is?

christophfroehlich: I'd use the current behavior, so basically just copy it in this common helper file. Using something with double-slash does not make a lot of sense, but would be valid I guess. I suggest that we change that in a follow-up PR, and only on rolling.

control_toolbox/include/control_toolbox/tf_utils.hpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2025, ros2_control developers
+// All rights reserved.
+//
+// Software License Agreement (BSD License 2.0)
```
christophfroehlich: I think that diff_drive_controller has the Apache-2 license, why have you chosen BSD here?

control_toolbox/test/tf_utils_tests.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2025, ros2_control developers
+// All rights reserved.
+//
+// Software License Agreement (BSD License 2.0)
```
christophfroehlich: same here

christophfroehlich (commented):
I'm fine with the current state, just consider my comment about the chosen license.

control_toolbox/include/control_toolbox/tf_utils.hpp
kuralme: There was no particular reason. I just intended to match the other ones in the toolbox since most of them use BSD.

PR #536, opened 2025-10-30
pid terms modification

aalmrad:
Modified the pid terms in the pid_ros.cpp file after receiving build errors.


aalmrad:
The modifications were tested only in Jazzy

aalmrad:
@christophfroehlich , jazzy branch of control_msgs solved the issue. Thanks.

PR #547, opened 2025-11-19
Add string() methods to get the printable information

saikishor:


control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -146,6 +146,13 @@ struct AntiWindupStrategy
               << "\terror_deadband: " << error_deadband << std::endl;
   }
 
+  std::string string() const
```
christophfroehlich, line 140: puh, this gets confusing with to_string() and std::string(). 
Can we at least reuse this in print()?

control_toolbox/include/control_toolbox/pid.hpp
```diff
@@ -326,6 +333,13 @@ class Pid
       antiwindup_strat_.print();
     }
 
+    std::string string() const
```
christophfroehlich, line 322: same here?

christophfroehlich (changes requested):

control_toolbox/include/control_toolbox/pid.hpp
saikishor: Sure

christophfroehlich (approved):

PR #549, opened 2025-11-29
Fixed Rate limiter's Asymmetric Initialization

thedevmystic:
Hello! Respected Maintainers and Reviewers!
Again, this is Surya!

This PR addresses the [issue #2029](https://github.com/ros-controls/ros2_controllers/issues/2029) opened in [ros2_controllers](https://github.com/ros-controls/ros2_controllers).

It fixes,
The promise of `description: "Maximum deceleration in reverse direction. If not set, -max_deceleration will be used."` given in [diff_drive_controller](https://github.com/ros-controls/ros2_controllers/blob/ec198b1d98ccb9a5271fa654396e18086aaa72cd/diff_drive_controller/src/diff_drive_controller_parameter.yaml#L157).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,7 @@ repos:
 
   # Python hooks
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.21.0
+    rev: v3.21.2
     hooks:
     -   id: pyupgrade
         args: [--py36-plus]
@@ -49,7 +49,7 @@ repos:
       args: ["--ignore=D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404"]
 
   - repo: https://github.com/psf/black
-    rev: 25.9.0
+    rev: 25.11.0
     hooks:
       - id: black
         args: ["--line-length=99"]
@@ -62,7 +62,7 @@ repos:
 
   # CPP hooks
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v21.1.2
+    rev: v21.1.6
     hooks:
       - id: clang-format
         args: ['-fallback-style=none', '-i']
@@ -136,7 +136,7 @@ repos:
         exclude: CHANGELOG\.rst|\.(svg|pyc)$
 
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.34.1
+    rev: 0.35.0
     hooks:
       - id: check-github-workflows
         args: ["--verbose"]
--- a/control_toolbox/CHANGELOG.rst
+++ b/control_toolbox/CHANGELOG.rst
@@ -2,6 +2,13 @@
 Changelog for package control_toolbox
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+6.1.0 (2025-12-01)
+------------------
+* Exponantial filter refactoring (`#493 <https://github.com/ros-controls/control_toolbox/issues/493>`_)
+* Replace deprecated rclcpp::spin_some() (`#541 <https://github.com/ros-controls/control_toolbox/issues/541>`_)
+* No need for the specialization of update() (`#509 <https://github.com/ros-controls/control_toolbox/issues/509>`_)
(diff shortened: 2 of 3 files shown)


control_toolbox/include/control_toolbox/rate_limiter.hpp
```diff
@@ -207,15 +207,19 @@ void RateLimiter<T>::set_params(
...
-      max_first_derivative_neg_ = max_first_derivative_pos_;
+      max_first_derivative_neg_ = -max_first_derivative_pos_;
     }
     if (std::isnan(min_first_derivative_pos_))
     {
-      min_first_derivative_pos_ = min_first_derivative_neg_;
+      min_first_derivative_pos_ = -min_first_derivative_neg_;
     }
-    if (has_first_derivative_limits_ && min_first_derivative_pos_ > max_first_derivative_neg_)
+    if (has_first_derivative_limits_ && asym_given &&
```
thedevmystic, line 221: I didn't removed it, instead I patched it.

But I suggest this is redundant. As we already has set the `min_first_derivative_pos_` and, `max_first_derivative_neg_`.

This is just my opinion, and I welcome yours too.

christophfroehlich, line 221:
```suggestion
    if (asym_given &&
```
Isn't this redundant, too? (L208)

christophfroehlich: I don't think the new criterion is necessary?

christophfroehlich (changes requested):
Can you please add tests for this?

thedevmystic:
I think it is necessary, because without this we got failure in tests, where failure shouldn't be expected. Like,

**Without `asym_given`:**

```log
/home/runner/work/control_toolbox/control_toolbox/.work/target_ws/src/control_toolbox/control_toolbox/test/rate_limiter.cpp:23
    Expected: control_toolbox::RateLimiter limiter( -1.0, std::numeric_limits<double>::quiet_NaN(), -1.0, 1.0, std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN(), -1.0, 1.0) doesn't throw an exception.
      Actual: it throws.
    /home/runner/work/control_toolbox/control_toolbox/.work/target_ws/src/control_toolbox/control_toolbox/test/rate_limiter.cpp:28
    Expected: control_toolbox::RateLimiter limiter( std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN(), -1.0, 1.0, std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN(), -1.0, 1.0) doesn't throw an exception.
      Actual: it throws.
    /home/runner/work/control_toolbox/control_toolbox/.work/target_ws/src/control_toolbox/control_toolbox/test/rate_limiter.cpp:33
    Expected: control_toolbox::RateLimiter limiter( std::numeric_limits<double>::quiet_NaN(), 1.0, -1.0, 1.0, std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN(), -1.0, 1.0) doesn't throw an exception.
      Actual: it throws.

And so on...
```

**Test Summary:**

```log
Summary: 90 tests, 0 errors, 6 failures, 0 skipped
```

**The reason:**

```log
>>>
- control_toolbox.RateLimiterTest testValueLimits
  <<< failure message
    unknown file
    C++ exception with description "Invalid first derivative limits" thrown in the test body.
```

This arises because we are using asymmetric limits, but the validation expects symmetric (This is my assumption).

The validation check itself seems logically flawed. So, I added the `asym_given` check. After this check all tests passed.

I also think, I'm just delegating the error. But I don't have any other plan under my sleeves. So, if you have any ideas or plans, please share it with me.

Thanks, @christophfroehlich!

thedevmystic:
It was opened for more than a month without any follow-up.
So, when I was clearing repositories that I don't use anymore, I deleted the head repository too.

If you want I can open the PR again. But the problem that I was encountering is still given above, if you can help please help!

PR #671, opened 2026-09-13
Fix RateLimiter default for max_first_derivative_neg

### LINKED ISSUE #2029: `max_deceleration_reverse` parameter does not default to `max_deceleration`
It's promised here:
https://github.com/ros-controls/ros2_controllers/blob/ec198b1d98ccb9a5271fa654396e18086aaa72cd/diff_drive_controller/src/diff_drive_controller_parameter.yaml#L157

The first nan check happens here I think:

https://github.com/ros-controls/control_toolbox/blob/33843ee427db3bf12d159995754580e2d5387066/control_toolbox/include/control_toolbox/rate_limiter.hpp#L210..L213
```c++
    if (std::isnan(max_first_derivative_neg_))
    {
      max_first_derivative_neg_ = max_first_derivative_pos_;
    }
```

Which is not the implementation promised.

KR-Ravindra:
#### Problem

`diff_drive_controller` documents `max_deceleration_reverse` as "If not set, -max_deceleration will be used", but the `RateLimiter` fallback uses the forward *acceleration* instead. A robot with asymmetric acceleration/deceleration limits therefore brakes with the wrong limit while driving in reverse.

#### Root cause

`control_toolbox/include/control_toolbox/rate_limiter.hpp`, `set_params()`: when `max_first_derivative_neg` (the reverse deceleration limit) is NaN it is set to `max_first_derivative_pos` (the forward acceleration limit) rather than to `-min_first_derivative_pos` (the forward deceleration limit).

#### Fix

- Default `max_first_derivative_neg` to `-min_first_derivative_pos`; resolve `min_first_derivative_pos` first so the case where both are unset still yields symmetric limits, as before.
- Correct the doxygen note that claimed "symmetric limits are used" for these two parameters.
- Add a test with `min_first_derivative_pos = -2.0` and `max_first_derivative_neg` unset, checking the reverse-deceleration clamp, plus a check that the fully symmetric configuration is unchanged.

Note on the snippet in the issue: `-max_first_derivative_pos` would produce a negative *upper* bound (that is why the tests in #549 threw). The documented promise is the negated forward *deceleration*, which is what this PR implements.

#### How tested

Before (new test against the unchanged header):
```
control_toolbox/test/rate_limiter.cpp:306: Failure
Expected equality of these values:
  v
    Which is: -4.5
  -4.0
[  FAILED  ] RateLimiterTest.testFirstDerivativeLimitsAsymmetricDefaults
```

After:
```
[==========] 8 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 8 tests.
```

Fixes ros-controls/ros2_controllers#2029

### Is this user-facing behavior change?

Yes, for users who leave `max_deceleration_reverse` (`max_first_derivative_neg`) unset:

- If `max_acceleration_reverse != -max_acceleration` and both `max_deceleration` and `max_deceleration_reverse` are unset, reverse deceleration now uses `|max_acceleration_reverse|` instead of `|max_acceleration|`, because `max_deceleration` already falls back to `max_acceleration_reverse` in the existing code.
- If `max_deceleration` is set and `max_deceleration_reverse` is unset, reverse deceleration now uses `|max_deceleration|` (the documented behaviour) instead of `|max_acceleration|`.
- The fully unset (symmetric) configuration is unchanged; the second block of the new test pins it.

### Did you use Generative AI?

Yes. This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### Additional Information

- Previous attempt: #549 (closed; it negated `max_first_derivative_pos`, which produced a negative upper bound and made the tests throw).
- Tests were run locally with a standalone gtest build of `control_toolbox/test/rate_limiter.cpp` and formatting checked with clang-format 23.1.0 (the pre-commit pinned version) and the repo `.clang-format`, rather than a full `colcon test` / `pre-commit run`.
- No `doc/release_notes.rst` entry added because the file has no section for the current cycle; happy to add one if wanted.

## TODOs

To send us a pull request, please:

- [x] Fork the repository.
- [x] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [ ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [x] Commit to your fork using clear commit messages.
- [x] Send a pull request, answering any default questions in the pull request interface.
- [x] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.


KR-Ravindra:
Round 1 self-review.

Verified against `master`:
1. `max_first_derivative_neg` is the upper clamp bound for `v0 < 0` (`rate_limiter.hpp` L286) and is documented as "usually >= 0", so the old fallback to `max_first_derivative_pos` (an acceleration limit) was the wrong magnitude and `-max_first_derivative_pos` (the #549 variant) has the wrong sign, which is why #549's `testWrongParams` threw. `-min_first_derivative_pos` is the deceleration limit with the right sign.
2. Compiled the new test against the `master` header: fails with `v = -4.5 vs -4.0`; against this branch 8/8 pass; the 7 existing tests still pass against the new header.
3. clang-format 23.1.0 with the repo `.clang-format` produces no diff on either file.
4. Unticked the `colcon test` / `pre-commit run` box in the description since only a standalone gtest build and clang-format were run.

No duplicate: #2029 only cross-references the closed #549 (fork deleted). Marking ready.

